Issue #82, opened 2018-05-09
Minor feature request - verify schema integrity before or when building deployment container


moxious:
Say I have a schema.yaml like this:

```
properties:
  foo:
    type: string
    default: hi

required:
- bar
- baz
```

This schema doesn't make any sense, and the deployment container is guaranteed to fail when at deploy time the python check script sees that you're requiring a variable that wasn't specified and hence probably wasn't passed.

It'd be helpful if when I'm building my docker container (like so: https://github.com/GoogleCloudPlatform/marketplace-k8s-app-example/blob/master/nginx/Makefile#L31) that there be some validation steps/checks.

Lacking this, I'm usually going through a cycle of build/push container, run start.sh job, monitor kubernetes logs, to determine that what I put into the container never made sense in the first place.

PR #177, opened 2018-07-10
Tooling for static verification of deployer image

vcanaa:
The validation is bundled with the deployer image, so partners have access to it.


marketplace/deployer_util/run_tester.py
```diff
@@ -74,9 +75,18 @@ def main():
...
+        for status in containerStatuses:
+          pendingReason = deep_get(status, "state", "waiting", "reason")
+          if pendingReason == INVALID_IMAGE_NAME_ERROR:
+            print_logs(full_name, args.namespace)
+            raise Exception("ERROR Tester '{}' failed: {}".format(full_name, pendingReason))
+
       if time.time() - start_time > tester_timeout:
         print_logs(full_name, args.namespace)
-        raise Exception("ERROR Tester '{}' timeout".format(full_name))
+        raise Exception("ERROR Tester '{}' timeout: {}".format(full_name, pendingReason))
```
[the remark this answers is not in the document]
vcanaa: Error messages are best effort. If `pendingReason` is None, we still do the same.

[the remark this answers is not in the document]
huyhg: Understood. However, in the case the error message would be confusing, or worst making the error reporting code look like it having a bug: `timeout: None`. I think we should condition on that to improve the message.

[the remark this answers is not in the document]
vcanaa: That is a very strong argument indeed. Lets do timeout: Reason undefined.

marketplace/deployer_util/deploy_with_tests.sh
```diff
@@ -54,6 +54,8 @@ separate_tester_resources.py \
   --out_manifests "/data/resources.yaml" \
   --out_test_manifests "/data/tester.yaml"
 
+cat "/data/resources.yaml"
```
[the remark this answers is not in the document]
huyhg: I was actually thinking about the following output:

```
=== /data/resources.yaml ===
<cat content goes here>
============================
```

marketplace/deployer_util/validate.py
```diff
@@ -0,0 +1,185 @@
...
+  print("Metadata validation succeeded")
+
+def print_file(filename):
+  with open(filename, "r") as stream:
+    print(filename)
+    print(stream.read())
+
+
+def deep_find(o, key, found=[]):
+  """ Finds all values with the specified key in any level of the object """
```
huyhg: nit: rm space right after and right before `"""`.

marketplace/deployer_util/validate.py
```diff
@@ -0,0 +1,185 @@
...
+  
+  if metadata:  
+    validate_metadata(metadata, application)
+
+  schema = load_yaml("/data/schema.yaml")
+  validate_images(schema, all_resources)
+
+
+def validate_images(schema, resources):
+  declared_images = set([deep_get(param, "default") for key, param in schema["properties"].iteritems()
```
huyhg: nit: reformat this as follow for readability:

```py
declared_images = set([
    deep_get(param, "default")
    for key, param in schema["properties"].iteritems()
    if deep_get(param, GOOGLE_MARKETPLACE, "type") == "IMAGE"])
```

marketplace/deployer_util/validate.py
```diff
@@ -0,0 +1,185 @@
...
+  declared_images = set([deep_get(param, "default") for key, param in schema["properties"].iteritems() 
+    if deep_get(param, GOOGLE_MARKETPLACE, "type") == "IMAGE"])
+
+  used_images = []
+  for resource in resources:
+    used_images.extend(deep_find(resource, "image"))
+  used_images = set(used_images)
+
+  undeclared_images = [image for image in used_images 
+    if image not in declared_images]
```
huyhg: nit: 4 space indent, or line up with `[`

```py
undeclared_images = [image for image in used_images
                     if image not in declared_images]
```

marketplace/deployer_util/validate.py
```diff
@@ -0,0 +1,185 @@
...
+      print("  " + str(image))
+
+    raise ValidationException(error_message)
+
+  print("All {} used images are declared.".format(len(used_images)))
+
+  unused_images = [image for image in declared_images 
+    if image not in used_images]
+
+  if unused_images:
```
huyhg: this validation is not always desired. A partner has a use case in which no workloads use a declared image, but they would still want to declare it to be republished, and the value is actually contained in a CRD. Their application will create new workloads using the image from the CRD. Thus, the validation here is probably the set of extracted images must be a subset of the declared images.

We also want to do the following:
- Re-expand the manifests with new values for the images.
- Ensure that the extracted images also change (subset of the new values)
- This catches the case where the an image is declared and also hard coded to the default value.

This logic is non-trivial. I would suggest the following:
- Modify the unit test first to create this scenario. The current code should fail.
- Add a block of comment explaining what we're doing here.

marketplace/deployer_util/validate.py
```diff
@@ -0,0 +1,185 @@
...
+
+  for image in used_images:
+    print(image)
+
+  print("Images validation succeeded")
+
+
+def expand_and_load_resources():
+  """Expands the templates and charts so we have the final manifests manifests"""
+  print(Command("/bin/expand_config.py").output)
```
huyhg: A few things here:
- `expand_config` will fail if a required property doesn't have a value. For example, service account or some custom property like email address. These properties will need a test value in order to pass validation anyway, so we can use those test values.
- The test values can probably be supplied in the test `schema.yaml`, as `default` values?
- I'm thinking we should allow the partner to run this to validate their image as part of their Dockerfile. This is great because errors will be caught at build time instead of runtime. See #82.
- For this to happen, this cannot rely on modifying local file system. Thus, we should probably modify `expand_config` to make it into a library and use it here. The main output of `expand_config` is a dictionary of input parameters.
- `create_manifests.sh` is another beast..
-

marketplace/deployer_util/validate.py
```diff
@@ -0,0 +1,185 @@
...
+  validate_images(schema, all_resources)
+
+
+def validate_images(schema, resources):
+  declared_images = set([deep_get(param, "default") for key, param in schema["properties"].iteritems() 
+    if deep_get(param, GOOGLE_MARKETPLACE, "type") == "IMAGE"])
+
+  used_images = []
+  for resource in resources:
+    used_images.extend(deep_find(resource, "image"))
```
huyhg: Not entirely sure about blanketing all attributes with `image` names. Maybe we should handle known workload resource types only.
- Pod
- Deployment
- StatefulSet
- DaemonSet
- Job
- CronJob
- ReplicaSet
- ReplicationController

See [k8s doc](https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/). There isn't a landing page that lists all the workloads! Please look at the content tree on the LHS.

marketplace/deployer_util/validate.py
vcanaa: I don't follow after "We also want to do the following...". Lets sync offline.

marketplace/deployer_util/validate.py
vcanaa: Is this python guideline? If so, that is kinda wasteful.

marketplace/deployer_util/validate.py
vcanaa: This is tricky. I would opt for false fails rather than false successes.

a) If we are too specific, we run the risk of skipping images.
b) If we do a catch all, we run the risk of doing extra validation.

I prefer fixing a bug in (b) rather than (a). If (b) causes trouble to us, that will be a partner complaint. If (a) causes trouble, that will probably be complains from multiple customers.

PR #182, opened 2018-07-12
Adding the version to base deployer images

vcanaa:


publish_deployers.sh
```diff
@@ -0,0 +1,75 @@
...
+exit 0
+fi
+
+changes="$(git status --porcelain)"
+if [[ ! -z "$changes" ]]; then
+  echo "Make sure there are no pending changes"
+  exit 1
+fi
+
+make -B .build/marketplace/deployer/envsubst
```
huyhg: TK has introduced a new `dev` image. Can you please include this into script as well?

PR #191, opened 2018-07-26
Pull ubbagent from https endpoint (rather than ssh).

trironkk:


trironkk:
It won't work in the example app until https://github.com/GoogleCloudPlatform/marketplace-k8s-app-example/pull/98 is merged.

PR #193, opened 2018-07-30
Reparenting marketplace/dev image to be based on official gcloud image.

trironkk:
This pull simplifies the handling of kubectl credentials files (removing need to mount volumes) and improves Cloudbuild performance (`gcr.io/cloud-builders/gcloud` image is already cached).

### DIFF
--- a/marketplace/dev/Dockerfile
+++ b/marketplace/dev/Dockerfile
@@ -1,4 +1,4 @@
-FROM launcher.gcr.io/google/ubuntu16_04
+FROM gcr.io/cloud-builders/gcloud
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
         apt-transport-https \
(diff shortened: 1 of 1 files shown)


marketplace/dev/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM launcher.gcr.io/google/ubuntu16_04
+FROM gcr.io/cloud-builders/gcloud
```
huyhg: Since we intend the dev image to be used widely in development (i.e. in non-cloud build environments), I'm a little hesitant about this move to using cloud-build's specific image. This flags a tight integration with cloud-build more that we had hoped for.

Can you detail the specific reason that requires this change?

trironkk: This reparenting is cloudbuild specific, and you're right to call this out. :)

In my opinion, the benefit of this change (specific to the cloudbuild context) comes at the cost of having our `kubectl` and `gcloud` binaries in a custom location. These binaries remain on `$PATH`, so I don't foresee changes in calling semantic, inside or outside the context of cloudbuild.

Regarding longer term plan, I don't anticipate a need for having an dev image explicitly decoupled from cloudbuild, as `gcr.io/cloud-builders/gcloud`'s [Dockerfile](https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/gcloud)'s construction doesn't look unique to cloudbuild. If it becomes so in the future, we could address such a breakage by forking the definition of the cloudbuild Dockerfiles at that point.

PR #195, opened 2018-08-02
Add Travis for pylint and py tests

huyhg:
The argument against using Cloud Build for this is mainly the availability of the logs. With Travis, the contributors, internal and external, can quickly see the errors and fix them. It doesn't make sense for these failures to require a round trip so that someone with GCP access can look at what errors occur in the log.

Switching to Cloud Build is easy if it's viable. I'm following with Cloud Build to see if there is a way to allow the logs to be available publicly.


huyhg:
Alright, sorry for all the churns. This PR is good for review now.

I've reformatted all of our code as well!

Makefile
```diff
@@ -25,22 +25,59 @@ include marketplace.Makefile
...
+	  --exclude "data/vendor/" \
+	  --parallel \
+	  "/data" \
+	  || (echo ">>> Linting failed. Run 'make format/py' to fix your code."; exit 1)
+	@docker run --rm \
+	  -v $(PWD):/data:ro \
+	  --entrypoint python2 \
+	  testing/py \
+	  -m pyflakes "/data"
+	@$(call print_notice,No lint errors found.)
```
trironkk: What do you think about wrapping this in a script to keep Makefile thin?

PR #196, opened 2018-08-02
Add Makefile to testrunner

jprzychodzen:


testrunner.Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+makefile_dir := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
+include $(makefile_dir)/common.Makefile
+
+
+.build/testrunner: | .build
+	$(call print_target)
+	cd $(MARKETPLACE_TOOLS_PATH)/testrunner \
+	  && bazel build //runner:main \
+	  && rm -rf tmp \
```
huyhg: I think we should put temp in the `.build` tree just to not pollute the source tree. We can use `$@/tmp` instead.

jprzychodzen: We cannot use `$@/tmp` for a `tmp` dir, because we use `touch $@` as a mark file for a target, so `$@` is a file, not a directory.

Also, `tmp` is in `testrunner/.gitignore`, and those steps are just steps from `testrunner/README.md`.

How do you feel about leaving this as is? And how about using `.build/tmp`, maybe this is a best solution in this case?

huyhg: Let's use `.build/tmp/testrunner`.

Ideally building shouldn't modify source, even if the files are in gitignore, because we'll want to remove all build artifacts with `make clean` too. So I do want to try to find a way not to pollute source.

jprzychodzen: Done.

common.Makefile
```diff
@@ -28,6 +28,10 @@ MARKETPLACE_TOOLS_PATH ?= $(patsubst %/,%,$(dir $(realpath $(lastword $(MAKEFILE
 	mkdir -p .build
 
 
+.build/tmp: | .build
+	mkdir -p .build/tmp
```
huyhg: nit: use `"$@"`. Please update the target above too.

testrunner.Makefile
```diff
@@ -0,0 +1,20 @@
...
+include $(makefile_dir)/common.Makefile
+
+
+.build/testrunner: | .build/tmp
+	$(call print_target)
+	cd $(MARKETPLACE_TOOLS_PATH)/testrunner \
+	  && bazel build //runner:main \
+	  && rm -rf ../.build/tmp/testrunner \
+	  && mkdir -p ../.build/tmp/testrunner \
+	  && cp bazel-bin/runner/main ../.build/tmp/testrunner/testrunner \
```
huyhg: this relies on the fact that `.build` is one directory above `$MARKETPLACE_TOOLS_PATH`, which isn't always the case. Split the command at this line:
- `cd` and `bazel build`
- `cp $(MARKETPLACE_TOOLS_PATH)/testrunner/bazel-bin/runner/main .build/tmp/...` and `docker build`

common.Makefile
jprzychodzen: Done

testrunner.Makefile
jprzychodzen: I've done this changes. Could you please verify that this is what you expected?

I'm sending in current, but if something is missing, please ping me by email, I will change this.

Thanks!

PR #198, opened 2018-08-03
Make driver.sh use specific /dev image

jprzychodzen:


publish_deployers.sh
```diff
@@ -54,16 +57,12 @@ if [[ ! -z "$changes" ]]; then
   exit 1
 fi
 
+image_tag=$( get_image_tag )
```
huyhg: nit: remove spaces

scripts/driver/driver.sh
```diff
@@ -52,12 +55,14 @@ done
 [[ -z "$kubeconfig" ]] && kubeconfig="${KUBECONFIG:-$HOME/.kube}"
 [[ -z "$gcloudconfig" ]] && gcloudconfig="$HOME/.config/gcloud"
 
+image_tag=$( get_image_tag )
```
huyhg: nit: remove spaces

scripts/driver/driver.sh
```diff
@@ -52,12 +55,14 @@ done
...
+
 docker run \
     --volume "/var/run/docker.sock:/var/run/docker.sock:ro" \
     --volume "$kubeconfig:/root/mount/.kube:ro" \
     --volume "$gcloudconfig:/root/mount/.config/gcloud:ro" \
     --rm \
-    "gcr.io/cloud-marketplace-tools/k8s/dev" \
+    "gcr.io/cloud-marketplace-tools/k8s/dev:$image_tag" \
```
huyhg: This will break local dev workflow: `make app/verify` locally builds this image with `latest` tag.

By specifying an `image_tag` here, we have 2 failure modes:
- if the tag happens to be a valid release tag, docker will pull it from the public repo. It will ignore all the local modifications
- if the tag doesn't exist in the public repo, this will just fail plainly

trironkk: This is a good call out - I missed that we were modifying `driver.sh` in this PR, and agree we should adjust this.

I've been thinking about this vaguely, but only just now (after sync'ing with @huyhg) have a concrete idea of how we should move forward. Here's my take:
* Extract `common.sh` functionality into a tool ("`derive_image_tag`"?) with the following semantics:
  * If the repo is clean, print out the the current commit sha.
  * if the repo is dirty, print out "latest".
* Change marketplace.Makefile targets that build images to invoke the tool to...
  * Derive the desired image tags.
  * If the tag is latest, build it.
  * If the tag does not exist in `gcr.io/cloud-marketplace-tools/`, build it.
  * Otherwise, pull it.
* Add a `--image_tag` parameter to `start.sh` and `driver.sh` and specify it by invoking the aforementioned tool.
* Add a Cloud Build hook to publish images with tag derived by aforementioned tool whenever pull requests are merged into `master`.

I'm picking up CI/CD work for this repo now, and can take this on. @jprzychodzen, is this blocking you? `make app/verify` succeed for me from a clean checkout of https://github.com/GoogleCloudPlatform/click-to-deploy/pull/200 - if that's merged will you still be blocked?

trironkk: Heads up - I'm starting work on exactly these changes.

huyhg: Passing `--image_tag` to `start.sh` seems odd.

There are 2 use cases:
- Partners (and ourselves) who vendor in tools and use our Makefile. In this case, what you're describing above seems reasonable.
- Partners who only use our images directly. They shouldn't need to vendor in the tools repo and shouldn't need `start.sh` source. Instead, they can pin to the dev image and run `start.sh` directly from there. In this case, I would like for the them to not have to specify `--image_tag`.

jprzychodzen: @huyhg , what do you mean by partners that are using images directly? You are refering to use case when partners will write they own pipeline for testing images, using `gcr.io/cloud-marketplace-tools/k8s/dev` image?

How partner can specify (or it will be detected) which tag needs to be used? Because right now these images are backward incompatible.

Also, `make app/verify` does not build `gcr.io/cloud-marketplace-tools/k8s/dev`. Even `make .build/marketplace/deployer/envsubst` does not build `gcr.io/cloud-marketplace-tools/k8s/dev`.

@trironkk, with PR/200, after removing all images from my image repository, and running `make clean`, `make app/verify` fails for `cassandra`. It downloads `gcr.io/cloud-marketplace-tools/k8s/dev:latest` image, without rebuilding locally (and I think it shouldn't rebuild locally for make app/verify).

Regarding failing painfully - I think this is ok, if one is using a version that is not released (tagged) and images are not available in public repository. For `click-to-deploy` repository we need to specify `deplyer_envsubst` version of an image. Version of submodule repository should be the same. Right now we are failing silently, and this is much worse solution. And non-compatibility in backward way, as we are using `latest` right now.

I would say that not-tagged commit is unreleased and unsupported for a pipeline (from partner perspective). If one wants to hack over it to add own modifications, sure, feel free to do that, but there should be some kind of a warning (like painfully failing with some way to recover).

I agree there are two use cases, but I think these are:
1. `marketplace-k8s-app-tools` repository, developing of a framework
2. repository with application(s) - both partners and `click-to-deploy`. This should use tagged versions of `marketplace-k8s-app-tools` and rather should download images, without rebuilding.

The problem is how we can make both use cases compatible and convenient to develop.

jprzychodzen:
Closed as oboselete

PR #200, opened 2018-08-06
Rename scripts/ to mirror make app/* targets.

trironkk:


scripts/verify
```diff
@@ -0,0 +1,177 @@
...
+function clean_and_exit() {
+  message=$1
+  [[ -z "$message" ]] || echo "$message"
+
+  delete_namespace
+  exit 1
+}
+
+echo "INFO Initializes the deployer container which will deploy all the application components"
+/scripts/start_internal.sh \
```
huyhg: Can we update any remaining references in `verify` and `install` so that we can get rid of `driver_internal.sh` and `start_internal.sh` altogether? These are not part of the deployer APIs so I don't see a reason to keep them around.

scripts/verify.sh
```diff
@@ -59,7 +59,7 @@ docker run \
     --rm \
     "gcr.io/cloud-marketplace-tools/k8s/dev" \
     -- \
-    /scripts/driver_internal.sh \
+    /scripts/verify \
```
huyhg: Having pairs of the likes of `install` and `install.sh` under `scripts` is a bit confusing. Did we decide to follow up and inline this into the Makefile?

scripts/verify
trironkk: I thought I had, but missed this. Good catch!

trironkk:
I've added image tag derivation to this pull request, as pushing to `master` without it will trigger a backward incompatible image to be copied to `gcr.io/cloud-marketplace-tools/`.

scripts/derive_tag.sh
```diff
@@ -15,27 +15,26 @@
 # limitations under the License.
 
 set -e
-set -o pipefail
 
 for i in "$@"
```
huyhg: Kill?

scripts/derive_tag.sh
```diff
@@ -15,27 +15,26 @@
 # limitations under the License.
 
 set -e
-set -o pipefail
```
huyhg: let's just keep it--it's standard. Or lets combine them: `set -eo pipefail`

scripts/derive_tag.sh
```diff
@@ -15,27 +15,26 @@
...
   *)
     >&2 echo "Unrecognized flag: $i"
     exit 1
     ;;
 esac
 done
 
-[[ -z "$name" ]] && >&2 echo "--name required" && exit 1
-[[ -z "$namespace" ]] && >&2 echo "--namespace required" && exit 1
+if [[ ! -z "$(git status --porcelain)" ]]; then
```
huyhg: what's the workflow for local development where this is expected?

app.Makefile
```diff
@@ -49,82 +49,73 @@ app/build:: ;
 app/install:: app/build \
               .build/var/APP_DEPLOYER_IMAGE \
               .build/var/APP_PARAMETERS \
-              .build/var/MARKETPLACE_TOOLS_PATH
+              .build/var/HOME \
+              .build/var/MARKETPLACE_TOOLS_TAG
```
huyhg: This is a good var, thanks!

For local workflows, we need to actually depend on a target that optionally rebuild the dev image.

What do you think about the following, which covers both local workflow and external cases:
- `.build/marketplace/dev` target should be an additional prerequisite here.
- `.build/marketplace/dev` also depends on `MARKETPLACE_TOOLS_TAG`, so it gets rebuilt when the tag changes.
- `MARKETPLACE_TOOLS_TAG` should be computed to something like `local` or `dirty` if it's actually dirty.
- If the tag is `local`/`dirty`, `.build/marketplace/dev` runs `docker build`.
- Otherwise, `.build/marketplace/dev` tries to pull. If it fails to pull, it will build. There are 2 reasons for this: (1) if the commit has been pushed to master, the pull is probably unnecessary (since `docker run` would pull as well) but does no harm; (2) it needs to build unpushed commits.
- There's one caveat here: if the tag goes from `local` to `sha_*` due to the dirty changes being committed, the dev image would be rebuilt while it doesn't need to. That's probably OK.

Note that `app.Makefile` in this case would have to import `marketplace.Makefile`, which it previously didn't need to.

trironkk: For local workflows, I had something similar in mind...

* Partner top level Makefile (e.g. wordpress/Makefile) would append `.build/marketplace/dev` as a dependency to `app/build`.
* `.build/marketplace/dev` runs `docker build` and updates `MARKETPLACE_TOOLS_TAG`.

I didn't include it in this pull request because I wanted to think more about tagging semantics for local development `local` or `dirty` or something. Probably just `local` is sufficient, as I don't think it's necessary to push the dev image anywhere, but wanted to think about it a bit more.

What do you think about this approach?

trironkk: Note: This pattern will break down if a `.build/marketplace/*` image is referenced by a manifest applied to a Kubernetes cluster, as the images are not pushed during local development. This is not a blocker, as these images are not on the install or verify flows.

scripts/verify
```diff
@@ -0,0 +1,176 @@
...
+function clean_and_exit() {
+  message=$1
+  [[ -z "$message" ]] || echo "$message"
+
+  delete_namespace
+  exit 1
+}
+
+echo "INFO Initializes the deployer container which will deploy all the application components"
+/scripts/install \
```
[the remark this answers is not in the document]
trironkk: Done.

scripts/start_internal.sh
```diff
@@ -1,114 +0,0 @@
-#!/bin/bash
```
huyhg: Can you completely delete these `*_internal.sh` files before merging?

scripts/derive_tag.sh
```diff
@@ -14,28 +14,17 @@
...
+  >&2 echo "-----------------------------------------------------------"
+  echo -n ''
+  exit 1
+fi
 
-[[ -z "$name" ]] && >&2 echo "--name required" && exit 1
-[[ -z "$namespace" ]] && >&2 echo "--namespace required" && exit 1
-
-kubectl delete "application/$name" --namespace="$namespace" --ignore-not-found
+echo -n "sha_$(git rev-parse HEAD)"
```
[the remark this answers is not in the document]
huyhg: I think the main problem with this is that we can't have the submodule in a dirty state, which is useful while developing.

trironkk:
Created https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/issues/207 to track local development story.

Issue #207, opened 2018-08-21
Clean up local development workflow


trironkk:

trironkk:
Created from discussions in https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/200.

trironkk:
As a developer, I can customize the gcloud configuration location.

PR #208, opened 2018-08-22
Use applicationApiVersion, still supporting old name

huyhg:
Also include a few fixes to broken Makefiles, and a short-term fix for gcloud config overriding, as indicated in #207.


marketplace.Makefile
```diff
@@ -15,8 +16,8 @@ include $(makefile_dir)/common.Makefile
 .build/marketplace/dev: $(MARKETPLACE_TOOLS_PATH)/marketplace/deployer_util/* \
                         $(MARKETPLACE_TOOLS_PATH)/marketplace/dev/* \
                         $(MARKETPLACE_TOOLS_PATH)/scripts/* \
-                        $(MARKETPLACE_TOOLS_PATH)/scripts/driver/* \
```
trironkk: Oops. Hrm, I wonder how my stuff built locally.

Thanks for the fix!

PR #214, opened 2018-09-25
Make it easy to run dev container

huyhg:
Introduce docker_entrypoint.sh that executes user entered commands.
This allows arbitrary command, not just bash, to be specified with
docker run and avoids the need to specify --entrypoint.

Include a dev script in the container. User can (1) pull the
container, (2) copying the dev content into a local file, and (3)
run the dev script instead of docker run. This results in very
straight-forward commands like:

```
./dev /scripts/install --deployer='gcr.io/my_deployer'
```

Tweak Makefile to use the new dev script--much more simplified.
Users can continue to override environment variables such as
GCLOUD_CONFIG or KUBE_CONFIG, in both make and direct script execution.


huyhg:
https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/b581ad591aaa8fca0f5ab84c27d834844647c824/docs/building-deployer.md

Please take a peek. WIP, but this is how partners will use our tooling. Related to the PRs that I sent today and also related to your tiller work.

scripts/dev
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -eo pipefail
+
+KUBE_CONFIG=${KUBE_CONFIG:-$HOME/.kube}
+GCLOUD_CONFIG=${GCLOUD_CONFIG:-$HOME/.config/gcloud}
+EXTRA_DOCKER_PARAMS=${EXTRA_DOCKER_PARAMS:-}
+MARKETPLACE_TOOLS_TAG=${MARKETPLACE_TOOLS_TAG:-latest}
+MARKETPLACE_TOOLS_CONTAINER=${MARKETPLACE_TOOLS_CONTAINER:-gcr.io/cloud-marketplace-tools/k8s/dev}
+
```
trironkk: what do you think about the name `MARKETPLACE_TOOLS_IMAGE` instead?

app.Makefile
```diff
@@ -28,14 +28,20 @@ $(shell echo '$(APP_PARAMETERS)' '$(APP_TEST_PARAMETERS)' \
...
-EXTRA_DOCKER_PARAMS ?=
 
 .build/app: | .build
 	mkdir -p "$@"
 
 
+.build/app/dev: .build/var/MARKETPLACE_TOOLS_TAG \
+              | .build/app
+	$(call print_target)
+	docker run \
```
trironkk: Could you help me understand why we extract `scripts/dev` from the dev image, rather than running it directly?

trironkk: Chatted offline on this topic. The intent here is to not rely on the local filesystem implementation, as we move toward minimizing the build system requirements of this repository.

I would find it surprising that changes to `scripts/dev` are not reflected in `make app/install` and would prefer to run it from the local filesystem for that reason, but now that I know this I suppose it isn't of much consequence.

The target dependency semantics aren't really that clear to me (changes to `scripts/dev` or rebuilds of `gcr.io/cloud-marketplace-tools/k8s/dev` don't trigger `.build/app/dev` rebuilds), but I'm developing the opinion that we should spend effort on getting the next major release published and work off of docker images rather than figuring out clever make voodoo to make the development workflow simpler.

No action item here.

marketplace/dev/Dockerfile
```diff
@@ -31,7 +31,8 @@ RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - && \
 
 COPY marketplace/deployer_util/* /bin/
 COPY scripts/* /scripts/
-COPY marketplace/dev/setup_kubeconfig.sh /bin/setup_kubeconfig.sh
+COPY marketplace/dev/*.sh /bin/
```
trironkk: nit: Could we move `marketplace/dev/setup_config.sh ` to `marketplace/dev/bin/` and change this instruction to `COPY marketplace/dev/bin/* /bin/`?

marketplace/dev/setup_config.sh
```diff
@@ -1,13 +1,29 @@
 #!/bin/bash
```
trironkk: How did you get github to recognize this move?

huyhg: I didn't do anything special. Probably it was moved with little diff in a commit, hence it was recognized.

marketplace/dev/Dockerfile
huyhg: Done.

huyhg:
PTAL

marketplace/dev/setup_config.sh
trironkk: Nice. :)

scripts/dev
trironkk: Opposed?

marketplace/dev/Dockerfile
trironkk: Thanks!

scripts/dev
huyhg: Ah, missed this. Done. Thanks for the reminder!

huyhg:
Just a note that these are being merged into a dev branch, not master.

PR #217, opened 2018-09-25
Update doc

huyhg:


huyhg:
This is good enough for reviewing now.

docs/schema.md
```diff
@@ -0,0 +1,346 @@
...
+database:
+  image:
+    name: gcr.io/google/mysql:5.6
+replicas: 3
+```
+
+We can define the following schema:
+
+```yaml
+properties:
```
trironkk: i think we need an `application_api_version: v1beta1` here?

docs/schema.md
```diff
@@ -0,0 +1,346 @@
...
+All parameters defined in `schema.yaml` can be used in manifests.
+
+### Helm based deployer
+
+They can be referenced in helm charts just like a regular value from `values.yaml`.
+
+Example of `schema.yaml`
+
+```yaml
+application_api_version: v1beta
```
trironkk: i think it should be v1beta**1**.

docs/schema.md
```diff
@@ -0,0 +1,346 @@
...
+properties:
+  port:
+    type: string
```
trironkk: should this be `integer`?

docs/schema.md
```diff
@@ -0,0 +1,346 @@
...
+
+```yaml
+mysql:
+  image: <path to image>
+```
+
+the property can ve referenced in the schema file as `mysql.image`
+
+```yaml
+application_api_version: v1beta
```
trironkk: i think it should be v1beta**1**.

docs/schema.md
```diff
@@ -0,0 +1,346 @@
...
+
+- Dots can be used for referencing nested values. For example, if the app makes use of
+`values.yaml` like below:
+
+```yaml
+mysql:
+  image: <path to image>
+```
+
+the property can ve referenced in the schema file as `mysql.image`
```
trironkk: nit: s/ve/be/

docs/building-deployer.md
```diff
@@ -0,0 +1,75 @@
...
+  Follow instructions [here](building-deployer-envsubst.md).
+
+## Publishing a new version
+
+A new application version is submitted by pushing the corresponding deployer image
+to the staging repo.
+
+### Tags
+
+The deployer image **must** carry the primary track ID as its docker tag.
```
eshiroma: Is there a reason that "must" and "should" are in bold? Is it to call out requirements for the partner?

docs/building-deployer.md
```diff
@@ -0,0 +1,75 @@
...
+Marketplace uses the _latest_ image of that primary track ID tag when it looks
```
eshiroma: Only one image at a time can have a given docker tag, right? So is the "_latest_ image" in this case equivalent to "the image that (currently) has the primary track ID tag"? (at the time that Marketplace "looks" for new versions)
Maybe it'd be good to avoid using the word "latest" here given that it is a special Docker tag.

docs/building-deployer.md
huyhg: RFC 2119. I added a RFC 2119 required blob at the top of the files now.

docs/building-deployer.md
```diff
@@ -0,0 +1,75 @@
...
+for new versions of each track. The deployer image **should** also carry a uniquethat
```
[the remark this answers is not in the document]
huyhg: RFC 2119. I added a RFC 2119 required blob at the top of the files now.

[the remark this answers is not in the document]
huyhg: remove `that` 😄

docs/building-deployer.md
huyhg: Agreed. I reworded it to:

```
Marketplace uses the image last tagged with that primary track ID tag when it looks
```

docs/schema.md
```diff
@@ -0,0 +1,346 @@
...
+deploy the application.
+
+The information provided by the user for properties defined in `schema.yaml` is available at deployment time.
+
+The schema file must exist at the well known location `/data/schema.yaml` in the deployer image.
+
+This is a simple example of a schema.yaml file:
+
+```yaml
+application_api_version: v1beta
```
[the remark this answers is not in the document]
huyhg: Done.

docs/schema.md
huyhg: This is not meant to be a full schema. It's for demonstrating the relationship between `values.yaml` and `properties`. Add "abbreviated schema" to make it clearer.

docs/schema.md
huyhg: Done.

docs/schema.md
huyhg: Yes, done.

docs/schema.md
huyhg: Done.

docs/schema.md
huyhg: Done.

PR #220, opened 2018-09-28
More reliably find chart tars

huyhg:
The previous `ls -Al` implementation didn't work in my testing.


huyhg:
My helm deployer didn't have any test chart and it failed to verify.

I tested `ls -lA` on my local shell. It doesn't handle the case of empty directory reliably.

PR #221, opened 2018-10-01
Add new Helm deployer leveraging local-tiller for stronger integration.

trironkk:
Most of the files in this pull request are generated/copied. It should be easier to do this review on a per-commit basis.


marketplace/deployer_helm_tiller_base/bin/post-install.sh
```diff
@@ -0,0 +1,96 @@
...
+      --arg app_api_version "$app_api_version" \
+  > "$manifests_directory/$kind.json"
+  echo
+done
+
+find "$manifests_directory" -type f -size 0 -delete
+
+if [[ ! -z "$(ls -A "$manifests_directory")" ]]; then
+	cat "$manifests_directory"/*
+  kubectl apply -f "$manifests_directory"
```
huyhg: We shouldn't be using `kubectl apply` for helm created resources, since they weren't created with `kubectl apply` (I tried deploying a chart, there is no kubectl annotation). You should see the following warning in the log:

```
Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply
```

My recommendation here is to use kubectl patch to __exactly__ add the labels and owner references.

marketplace/deployer_helm_tiller_base/test/application.yaml
```diff
@@ -0,0 +1,44 @@
+{{- if .Values.template_mode }}
```
huyhg: Instead of an implicit value like this, we should introduce a new field in the schema. This is required by the helm-tiller deployer, but not by the UI:

```
properties:
  existingApplicationUid:
    type: string
    x-google-marketplace:
      type: EXISTING_APPLICATION_UID
```

The deployer will populate this with the right UID it finds. This line in the template should be rewritted as:

```
{{- if not .Values.existingApplicationUid }}
```

marketplace/deployer_helm_tiller_base/test/post-install.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: v1
+kind: ServiceAccount
```
huyhg: I'm concerned about the creation of these service accounts and roles/role bindings. We're building our RBAC hacks into the deployer.

Since this should run while the deployer is still alive, this should use the deployer's Service Account.

We should introduce (yet!) a new property that takes in the deployer's service account and use it here.

marketplace/deployer_helm_tiller_base/test/post-delete.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: v1
```
huyhg: Creation of these are more acceptable, as these objects are created not by the deployer but by the user (or server side tiller).

There is still a concern that we rely on `cluster-admin` role. The user might not have this permission (note that not only role creation but binding operations are also checked; somehow it works with IAM integration).

Since this only triggers when the user enters the CLI world, I think it's acceptable for now.

Probably the better way is to keep the deployer's service account around to run all hooks. (See my comment about alternate approach to owner referencing using container lifecycle).

marketplace/deployer_helm_tiller_base/bin/post-install.sh
```diff
@@ -0,0 +1,96 @@
...
+      --output=yaml \
+    | yaml2json \
+    | jq '.items[]
+            | {
+                apiVersion: .apiVersion,
+                kind: .kind,
+                metadata: .metadata
+              }
+            | del(.metadata.creationTimestamp)
+            | .metadata.ownerReferences +=
```
huyhg: An alternate approach to assigning owner references to all objects is to not do so and use a watcher container who's [terminating lifecycle hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) will call helm delete.

Essentially, we create a new deployment that is owned by the application. When the application is deleted, the deployment is also GC'ed. The pod will be signaled to terminate. The container terminating hook will run helm delete on the application.

Now there's a bit of a recursion going here when the user invokes helm delete, as the post-install hook will delete the application, which results in another helm delete operation from this mechanism. We just need to ignore the helm delete failure when this hook executes.

marketplace/deployer_helm_tiller_base/test/post-install.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: v1
```
huyhg: Requiring partners to supply these two hooks are really not ideal. We can automatically create a wrapper chart (either at docker build time or at run time). See [this example](https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/huyhg/doc/d/docs/building-deployer-helm.md#part-1b-recommended-creating-a-wrapper-chart) on how wrapper chart works.

There is one wrinkle: if the user tries to manipulate the release with helm directly, they will need access to the wrapper chart source. This encourages creating the wrapper chart at build time, so that the end user can just download the source code for it directly from the container.

The nice thing about helm chart is that in its `Chart.yaml` there is instruction on how to fetch a different version of the chart. So the main chart in this case should still have that declared properly, and the user after downloading the wrapper chart can still upgrade the wrapped main chart. Or they can just hand edit the thing.

marketplace/deployer_helm_tiller_base/test/application.yaml
trironkk: Created https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/224.

Will update when merged.

marketplace/deployer_helm_tiller_base/test/post-install.yaml
trironkk: Agreed that we should avoid referencing cluster-admin, but I think we can downscope permissions further than the deployer's service account. We only need to fetch applications and patch in component kinds.

Let's discuss this and the post-delete case offline.

marketplace/deployer_helm_tiller_base/test/post-install.yaml
trironkk: I have some hesitations about marketplace generating the wrapper chart automatically, but having the partner model their deployer as a wrapper chart itself is definitely cleaner.

Let's discuss if you feel strongly about marketplace generating the wrapper chart. :)

marketplace/deployer_helm_tiller_base/test/post-delete.yaml
trironkk: Let's discuss. :)

marketplace/deployer_helm_tiller_base/test/post-install.yaml
trironkk: Recapping offline discussion - a major reason for using the deployer service account here is to be consistent with disallowing partners from creating RBAC resources.

I have the impression that we disallow Partners creating RBAC resources that have privileges outside the application's namespace, but RBAC resources within the namespace are allowed.

marketplace/deployer_helm_tiller_base/bin/deploy.sh
```diff
@@ -0,0 +1,52 @@
...
+                 --values_file /data/values.yaml --values_dir /data/values)"
+    export NAME
+    export NAMESPACE
+  fi
+  patch_assembly_phase.sh --status="Failed"
+  exit $code
+}
+trap "handle_failure" EXIT
+
+/bin/expand_config.py
```
huyhg: needs merging

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,89 @@
...
+
+for chart in /data/chart/*; do
+  # Expand and apply the template for the Application resource.
+  # Note: This must be done out-of-band because the Application resource
+  # is created before the deployer is invoked, but helm does not handle
+  # pre-existing resources.
+  helm template \
+      --name="$NAME" \
+      --namespace="$NAMESPACE" \
+      --set "template_mode=true" \
```
huyhg: pending application uid integration right?

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,89 @@
...
+      --values=<(print_config.py \
+      --output=yaml) \
```
huyhg: nit: to previous line

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,89 @@
...
+  # it runs a local tiller process to perform template expansion and
+  # hook processing.
+  # Note: The local tiller process is configured with --storage=secret,
+  # which is recommended but not default behavior.
+  helm tiller run "$NAMESPACE" -- \
+    helm install \
+        --name="$NAME" \
+        --namespace="$NAMESPACE" \
+        --values=<(print_config.py \
+        --output=yaml) \
```
huyhg: nit: to previous line

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+appVersion: 0.7.1
+description: Marketplace Helm integration
+home: https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/
+keywords:
+- marketplace
+- helm
+maintainers:
+- email: cloud-marketplace-eng@google.copm
```
huyhg: not a public email alias. We probably have to set up one.

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/templates/post-delete.yaml
```diff
@@ -0,0 +1,30 @@
...
+        heritage: {{.Release.Service | quote }}
+        release: {{.Release.Name | quote }}
+        chart: "{{.Chart.Name}}-{{.Chart.Version}}"
+    spec:
+      serviceAccount: "{{.Values.service_account}}"
+      restartPolicy: Never
+      containers:
+      - name: post-delete-job
+        image: "{{.Values.image}}"
+        command:
```
huyhg: would recommend adding imagePullPolicy and expose that in values.yaml

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/templates/post-install.yaml
```diff
@@ -0,0 +1,30 @@
...
+        heritage: {{.Release.Service | quote }}
+        release: {{.Release.Name | quote }}
+        chart: "{{.Chart.Name}}-{{.Chart.Version}}"
+    spec:
+      serviceAccount: "{{.Values.service_account}}"
+      restartPolicy: Never
+      containers:
+      - name: post-install-job
+        image: "{{.Values.image}}"
+        command:
```
huyhg: would recommend adding imagePullPolicy and expose that in values.yaml

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/templates/post-install.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: Job
+metadata:
+  name: "{{.Release.Name}}-post-install"
+  labels:
+    heritage: {{.Release.Service | quote }}
+    release: {{.Release.Name | quote }}
+    chart: "{{.Chart.Name}}-{{.Chart.Version}}"
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-delete-policy": before-hook-creation
```
huyhg: do we really want to keep this job around post install?

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/templates/post-install.yaml
```diff
@@ -0,0 +1,30 @@
...
+    "helm.sh/hook-weight": "30"
+spec:
+  template:
+    metadata:
```
huyhg: remove this section--pod template doesn't have one

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/templates/post-delete.yaml
```diff
@@ -0,0 +1,30 @@
...
+    heritage: {{.Release.Service | quote }}
+    release: {{.Release.Name | quote }}
+    chart: "{{.Chart.Name}}-{{.Chart.Version}}"
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-delete-policy": hook-succeeded
+    "helm.sh/hook-weight": "30"
+spec:
+  template:
+    metadata:
```
huyhg: remove this section--pod template doesn't have one

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/templates/post-delete.yaml
```diff
@@ -0,0 +1,30 @@
...
+      name: "{{.Release.Name}}"
+      labels:
+        heritage: {{.Release.Service | quote }}
+        release: {{.Release.Name | quote }}
+        chart: "{{.Chart.Name}}-{{.Chart.Version}}"
+    spec:
+      serviceAccount: "{{.Values.service_account}}"
```
huyhg: do we want to add a backoffLimit = 0 for the jobs?

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+appVersion: 0.7.1
```
huyhg: we probably need a way to ensure this matches the release tag

marketplace/deployer_helm_tiller_base/test/chartmuseum-marketplace/requirements.yaml
```diff
@@ -0,0 +1,7 @@
+dependencies:
+- name: marketplace-k8s-app-tools-helm-hooks
+  version: 0.0.2
+  repository: gs://cloud-marketplace-tools-charts/
+- name: chartmuseum
```
huyhg: wait, we don't include a copy of this. Does that mean that `helm install`--i.e. the deployer--downloads it at runtime?

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,89 @@
...
+                  uid: $app_uid,
+                  blockOwnerDeletion: true
+                }
+              ]
+         ' \
+      --arg name "$NAME" \
+      --arg namespace "$namespace" \
+      --arg app_uid "$app_uid" \
+      --arg app_api_version "$app_api_version" \
+    | kubectl apply -f -
```
huyhg: use patch

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/Chart.yaml
trironkk: Not really sure what the right long term solution is, but I'll start with a small list (trironkk@google.com). If it gets noisy I'll change it.

marketplace/deployer_helm_tiller_base/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+RUN mkdir -p /bin/helm-downloaded \
+    && wget -q -O /bin/helm-downloaded/helm.tar.gz \
+        https://storage.googleapis.com/kubernetes-helm/helm-v2.10.0-linux-amd64.tar.gz \
+    && tar -zxvf /bin/helm-downloaded/helm.tar.gz -C /bin/helm-downloaded \
+    && mv /bin/helm-downloaded/linux-amd64/helm /bin/ \
+    && rm -rf /bin/helm-downloaded \
+    && helm init --client-only \
+    && helm plugin install https://github.com/rimusz/helm-tiller \
+    && helm plugin install https://github.com/nouney/helm-gcs \
```
huyhg: remove

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,108 @@
...
+for chart in /data/chart/*; do
+  # Expand and apply the template for the Application resource.
+  # Note: This must be done out-of-band because the Application resource
+  # is created before the deployer is invoked, but helm does not handle
+  # pre-existing resources.
+  helm template \
+      --name="$NAME" \
+      --namespace="$NAMESPACE" \
+      --values=<(print_config.py --output=yaml) \
+			--set "$(cat /data/schema.yaml \
```
huyhg: indent?

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,108 @@
...
+                 | yaml2json \
+                 | jq -r '.properties
```
huyhg: Let's limit the use of `jq`, hard to read to most and not testable. Can you update `print_config.py` to add support for a new flag `--key`. It should be similar to `--param`, except it prints the key instead of the value.

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,108 @@
...
+                "app.kubernetes.io/namespace": $namespace
+              }
+            }
+          }' \
+          --arg name "$NAME" \
+          --arg namespace "$NAMESPACE" \
+          --arg app_uid "$app_uid" \
+          --arg app_api_version "$app_api_version")"
+
+	kubectl get secrets \
```
huyhg: indentation for this block

marketplace/deployer_helm_tiller_base/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    && helm repo add cloud-marketplace-tools gs://cloud-marketplace-tools-charts/
+
+COPY marketplace/deployer_util/* /bin/
+COPY marketplace/deployer_helm_tiller_base/bin/* /bin/
+
+RUN mkdir -p /data/chart
+
+ENTRYPOINT ["/bin/bash", "/bin/deploy.sh"]
```
huyhg: I'd like to introduce an `entrypoint.sh` similar to the `dev` image that will do `exec "$@"` at the end. It's a nice paradigm to follow, to avoid having to change entrypoint of the docker container--only args need specifying. 

As long as that `entrypoint.sh` does do any required setup, we can add it now. It's backward compatible with the way our test run modifies the entrypoint.

Would suggest at least dropping `/bin/bash`, having this only as `ENTRYPOINT ["/bin/deploy.sh"]`. `/bin/bash` doesn't add much value; I find it more confusing to use, like one would wonder why specifying `echo "Hello world"` to the docker run command doesn't work.

marketplace/deployer_helm_tiller_base/test/schema.yaml
```diff
@@ -0,0 +1,34 @@
...
+    x-google-marketplace:
+      type: NAMESPACE
+  application_uid:
+    type: string
+    x-google-marketplace:
+      type: APPLICATION_UID
+  marketplace-k8s-app-tools-helm-hooks.image:
+    type: string
+    x-google-marketplace:
+      type: IMAGE
```
huyhg: Can we have a special property for the deployer image? I think the hook should just use the deployer image.

marketplace/deployer_helm_tiller_base/bin/deploy_with_tests.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
huyhg: It only makes sense to have these files in the deployer image if it's also used for running the hooks. I think that's a good idea. See my comment below about supporting a property for passing back the deployer image to the solution.

marketplace/deployer_helm_tiller_base/test/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG MARKETPLACE_TOOLS_TAG
+FROM gcr.io/cloud-marketplace-tools/k8s/deployer_helm_tiller:$MARKETPLACE_TOOLS_TAG
+
+COPY marketplace/deployer_helm_tiller_base/test/schema.yaml /data
+
+COPY marketplace/deployer_helm_tiller_base/test/chartmuseum-marketplace /tmp/chartmuseum-marketplace
+
+RUN mkdir -p /data/chart \
+ && cd /data/chart \
+ && helm dependency build /tmp/chartmuseum-marketplace \
```
huyhg: I thought more about this and we really shouldn't dictate how the partner builds the helm chart. This command shouldn't be part of ONBUILD.

The CI test can run `helm dependency build` to build out the chart directory. The test would then also run docker build, etc.

For future follow up.

marketplace/deployer_helm_tiller_base/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN mkdir -p /bin/helm-downloaded \
+    && wget -q -O /bin/helm-downloaded/helm.tar.gz \
+        https://storage.googleapis.com/kubernetes-helm/helm-v2.10.0-linux-amd64.tar.gz \
+    && tar -zxvf /bin/helm-downloaded/helm.tar.gz -C /bin/helm-downloaded \
+    && mv /bin/helm-downloaded/linux-amd64/helm /bin/ \
+    && rm -rf /bin/helm-downloaded \
+    && helm init --client-only \
+    && helm plugin install https://github.com/rimusz/helm-tiller \
+    && helm plugin install https://github.com/nouney/helm-gcs \
+    && helm repo add cloud-marketplace-tools gs://cloud-marketplace-tools-charts/
```
huyhg: https://github.com/helm/helm/blob/master/docs/chart_repository.md#google-cloud-storage

This can simply be: https://cloud-marketplace-tools.storage.googleapis.com/charts

marketplace/deployer_helm_tiller_base/test/chartmuseum-marketplace/requirements.yaml
```diff
@@ -0,0 +1,7 @@
+dependencies:
+- name: marketplace-k8s-app-tools-helm-hooks
+  version: 0.0.2
+  repository: gs://cloud-marketplace-tools-charts/
```
huyhg: use https link like above

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
huyhg: Would suggest renaming the chart to simply `application` and moving this to this directory:

```
/marketplace/charts/application
```

This directory essentially is the source for https://cloud-marketplace-tools.storage.googleapis.com/charts

marketplace/deployer_helm_tiller_base/Dockerfile
trironkk: Done.

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
trironkk: Done.

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
trironkk: I'd prefer to do this as a follow up - it's not partner facing.

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
trironkk: Done.

marketplace/deployer_helm_tiller_base/Dockerfile
trironkk: I'd prefer to do this as a follow up - it's not partner facing.

marketplace/deployer_helm_tiller_base/test/schema.yaml
trironkk: Maybe this isn't the right use of the deployer image, and we should bake/publish a different one?

marketplace/deployer_helm_tiller_base/Dockerfile
trironkk: Done.

marketplace/deployer_helm_tiller_base/test/chartmuseum-marketplace/requirements.yaml
trironkk: Done.

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/Chart.yaml
trironkk: Moved chart definitions to `marketplace/charts` and renamed to `marketplace-hooks`.

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
trironkk: Tracked in https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/issues/237.

marketplace/deployer_helm_tiller_base/marketplace-k8s-app-tools-helm-hooks/Chart.yaml
```diff
@@ -6,6 +6,6 @@ keywords:
 - marketplace
 - helm
 maintainers:
-- email: cloud-marketplace-eng@google.copm
-  name: google-cloud-marketplace
+- email: trironkk@google.com
```
huyhg: Please use cloud-marketplace-tools@google.com. Name can be `Marketplace Tools`.

trironkk: Done!

marketplace/charts/marketplace-hooks/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
huyhg: The reason why I don't want to call this chart "hooks" is because we might extend it to include more in the future, for example a controller. The idea here is this chart supports the application to run in GCP Marketplace.

How about one of the following:
- marketplace-app-support
- marketplace-application
- application-support

marketplace/deployer_helm_tiller_base/test/schema.yaml
huyhg: The issue with having a different image is that the partner has to copy that image into their staging repo. It's not terrible for our partners, but it doesn't seem necessary.

marketplace.Makefile
```diff
@@ -81,36 +81,26 @@ include $(makefile_dir)/var.Makefile
...
+.build/marketplace/charts/marketplace-hooks: \
+		.build/var/CHART_BUCKET \
+		.build/var/MARKETPLACE_HOOKS_VERSION \
+		marketplace/charts/Dockerfile \
+		$(shell find marketplace/charts/ -type f) \
+		$(MARKETPLACE_TOOLS_PATH)/marketplace/charts/marketplace-hooks/* \
+		| .build/marketplace/charts
 	$(call print_target)
 	cd $(MARKETPLACE_TOOLS_PATH) \
 	&& docker build \
```
huyhg: so we're building a docker image to get a side effect of publishing the chart?

marketplace.Makefile
```diff
@@ -5,6 +5,7 @@ __MARKETPLACE_MAKEFILE__ := included
 COMMIT ?= $(shell git rev-parse HEAD | fold -w 12 | head -n 1)
 
 makefile_dir := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
+include $(makefile_dir)/gcloud.Makefile
```
huyhg: this was intended for applications only. Why did we need to add it here?

marketplace/charts/marketplace-hooks/values.yaml
```diff
@@ -0,0 +1,2 @@
+imagePullPolicy: IfNotPresent
```
huyhg: missing `image` property here. A note that many helm charts follow convention of separate the image name into components: registry (gcr.io), repository (google/deployer), and tag.

marketplace.Makefile
```diff
@@ -62,6 +63,46 @@ include $(makefile_dir)/var.Makefile
...
+.build/marketplace/deployer/marketplace-k8s-app-tools-helm-hooks:
+	mkdir -p "$@"
+
+
+.build/marketplace/charts: | .build/marketplace
+	mkdir -p "$@"
+
+CHART_BUCKET = marketplace-k8s-app-tools-charts
+MARKETPLACE_HOOKS_VERSION = 0.0.1
+.build/marketplace/charts/marketplace-hooks: \
```
huyhg: these targets are intended to build these artifacts __locally only__. IIUIC, this pushes the charts to the release GCS.

marketplace/deployer_helm_tiller_base/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+              >&2 echo "Missing schema.yaml." ; \
+              find /tmp/onbuild ; \
+              exit 1 ; \
+            fi
+
+ONBUILD RUN export CHART_NAME="$(find /tmp/onbuild -maxdepth 1 -mindepth 1 -type d -printf '%P\n')" \
+            && cp /tmp/onbuild/schema.yaml /data/schema.yaml \
+            && mkdir -p /data/chart \
+            && cd /data/chart \
+            && helm package --dependency-update "/tmp/onbuild/$CHART_NAME" --destination /data/chart/ \
```
huyhg: can we drop `--dependency-update`? I don't think we should update dependencies automatically. To compare, we never want to update vendored go libraries automatically. It should be done manually and in a separate commit.

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,108 @@
...
+  helm tiller run "$NAMESPACE" -- \
+    helm install \
+        --name="$NAME" \
+        --namespace="$NAMESPACE" \
+        --values=<(print_config.py --output=yaml) \
+        "$chart"
+
+  # Establish an ownerReference back to the Application resource, so that
+  # the helm release will be cleaned up when the Application is deleted.
+  # Note: This is fragile, as any time that tiller updates the release
```
huyhg: hmm, shouldn't this then be in the post install hook? It should be able to query the application object using labels.

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,108 @@
...
+      --set "$(cat /data/schema.yaml \
+                 | yaml2json \
+                 | jq -r '.properties
+                            | to_entries
+                            | .[]
+                            | select(.value["x-google-marketplace"].type == "APPLICATION_UID")
+                            | .key')=" \
+      "$chart" \
+    | yaml2json \
+    | jq 'select( .kind == "Application" )' \
```
huyhg: (perhaps a follow up): Set proper labels for the application

trironkk: I thought that application labels ought to be set by `scripts/install` (or our backend)?

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
trironkk: Unfortunately it can't - touching the secret from within the hook causes a failure.

marketplace.Makefile
trironkk: We didn't - my mistake!

marketplace.Makefile
trironkk: Yes - in this way we can standardize the environment in which we build/publish charts.

marketplace/charts/marketplace-hooks/Chart.yaml
trironkk: Sync'd offline - moving forward with `marketplace-integration`.

marketplace/charts/marketplace-hooks/values.yaml
trironkk: i don't think it's necessary to provide a default for `image` property.

in a follow up pull request, i'd like to set it to the base helm tiller deployer image, but that will require a little more leg work on our release process to associate chart releases with deployer image releases.

marketplace/deployer_helm_tiller_base/Dockerfile
trironkk: Done!

marketplace/deployer_helm_tiller_base/bin/deploy_with_tests.sh
trironkk: see other comment for discussion.

marketplace.Makefile
trironkk: You're correct.

There more work to be done to standardize our release process for charts, but I'd like to tackle that in a followup.

marketplace.Makefile
huyhg: Let's not add release scripts to this makefile. This is intended for local development only. Note that non of our make targets currently update production artifacts, and it's a good thing. Production artifacts are updated by cloudbuild. Let's move this there.

marketplace/charts/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+ARG CHART_BUCKET
+ENV CHART_BUCKET $CHART_BUCKET
+
+ARG VERSION
+ENV VERSION $VERSION
+
+COPY . /charts
+
+RUN find /charts \
```
huyhg: I have mixed feelings about this. Why don't we have this as a script inside `dev`, and just run it to push a new version of the chart?

Note that we shouldn't build charts on every commit, since we need to make a versioned tgz file. Ideally, we should only build charts on new release tags. Would suggest moving all building and releasing charts to a follow up.

marketplace/deployer_helm_tiller_base/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+ENTRYPOINT ["/bin/bash", "/bin/deploy.sh"]
+
+ONBUILD COPY . /tmp/onbuild
+
+ONBUILD ARG HELM_DEPENDENCY_BUILD=false
+ONBUILD ENV HELM_DEPENDENCY_BUILD $HELM_DEPENDENCY_BUILD
+
+# Partner directory structure assertions.
+ONBUILD RUN if [ "$(find /tmp/onbuild -maxdepth 1 -mindepth 1 -type d -printf '\n' | wc -l)" -ne 1 ]; then \
```
huyhg: can we make sure that the directory that this expects is the same as in the [guide](https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/master/docs/building-deployer-helm.md#download-the-helm-chart). We should keep the convention consistent so that we can just point partners to the doc.

The internal of the tiller deployer can be different from the non-tiller one (that's an advantage of the onbuild paradigm that we discussed).

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
huyhg: But it would be overwritten by the one from the chart. I'm sure about the behavior of the current deployer. Likely that it doesn't alter the `Application` manifest. We don't have to do it here, but it's a good consistent thing to do, since we're already modifying the labels of the helm resources.

marketplace/deployer_helm_tiller_base/bin/deploy_with_tests.sh
huyhg: (Replying here as I can't reply on the other thread--seems like the commit was gone)

Note that we have a requirement that all images have to come from `marketplace.gcr.io`. Pointing this to a base deployer requires us publishing that to `marketplace.gcr.io` separately. I think it's more work for no benefits. Since the version of the deployer image is guaranteed to be consistent with that of the application helm chart, it seems more straight-forward and fool-proof to use that image.

The caveat is that it requires some boiler-plate in `schema.yaml` to plumb the deployer image through.

marketplace/deployer_helm_tiller_base/Dockerfile
trironkk: Done!

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
trironkk: Makes sense and agreed. Will take as a follow up.

marketplace.Makefile
trironkk: I'll stage chart release process changes separately. :)

marketplace/charts/Dockerfile
trironkk: Let's continue this discussion in the follow up pull request. :)

marketplace/deployer_helm_tiller_base/bin/deploy_with_tests.sh
trironkk: (Oops - pull rebase'd from `master`. Maybe I ought not to do that. Here's that comment for posterity:

> i don't think it's necessary to provide a default for `image` property.
> 
> in a follow up pull request, i'd like to set it to the base helm tiller deployer image, but that will require a little more leg work on our release process to associate chart releases with deployer image releases.)

A couple things aren't clear to me - let's sync offline.

trironkk:
Follow ups:
- [ ] Support for original helm deployer
- [ ] print_config.py support for extracting keys
- [ ] Chart Release Process
- [ ] Application Labels
- [ ] Add entrypoint.sh to Dockerfile for deployers.
- [ ] Example repo refactor.

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
```diff
@@ -0,0 +1,102 @@
...
+for chart in /data/chart/*; do
+  # Expand and apply the template for the Application resource.
+  # Note: This must be done out-of-band because the Application resource
+  # is created before the deployer is invoked, but helm does not handle
+  # pre-existing resources.
+  helm template \
+      --name="$NAME" \
+      --namespace="$NAMESPACE" \
+      --values=<(print_config.py --output=yaml) \
+      --set "$(print_config.py --xtype APPLICATION_UID --key true)=" \
```
huyhg: Nit: Oversight from my part: `--key` argument should have an action of recording true when specified, so that we don't need to specify `true` here. Not blocking.

marketplace/deployer_helm_tiller_base/bin/deploy_with_tests.sh
huyhg: Resolved offline. For the record, we want to use the (application) deployer image as the container to run the hooks.

marketplace/deployer_helm_tiller_base/onbuild/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+              find /tmp/onbuild ; \
+              exit 1 ; \
+            fi
+
+ONBUILD RUN export CHART_NAME="$(find /tmp/onbuild/chart -maxdepth 1 -mindepth 1 -type d -printf '%P\n')" \
+            && cp /tmp/onbuild/schema.yaml /data/schema.yaml \
+            && mkdir -p /data/chart \
+            && cd /data/chart \
+            && if [ "$HELM_DEPENDENCY_BUILD" = "true" ]; then \
+                 helm dependency update "/tmp/onbuild/chart/$CHART_NAME" ; \
```
huyhg: `s/update/build/`

testing/marketplace/deployer/helm_tiller_onbuild/helm-dependency-build/schema.yaml
```diff
@@ -0,0 +1,34 @@
+properties:
```
huyhg: If you could, please change the test directories to mirror the directory of the base deployer, instead of the make target. The goal is to easily find the source that this is testing, similar to `java` and `javatests` convention.

```
testing/marketplace/deployer_helm_tiller_base/onbuild/helm-dependency-build
testing/marketplace/deployer_helm_tiller_base/onbuild/standard
```

testing.Makefile
```diff
@@ -0,0 +1,46 @@
...
+		.build/marketplace/dev \
+		.build/var/MARKETPLACE_TOOLS_TAG \
+		.build/var/REGISTRY \
+		$(shell find testing/deployer_helm_tiller_base/helm-dependency-build -type f) \
+    | .testing/marketplace/deployer/helm_tiller_onbuild
+	$(call print_target)
+	TEST_ID=$(TEST_ID) \
+	REGISTRY=$(REGISTRY) \
+	MARKETPLACE_TOOLS_TAG=$(MARKETPLACE_TOOLS_TAG) \
+	  ./testing/marketplace/deployer/helm_tiller_onbuild/helm-dependency-build/test
```
huyhg: `touch "$@"`

testing.Makefile
```diff
@@ -0,0 +1,46 @@
...
+		.build/marketplace/dev \
+		.build/var/MARKETPLACE_TOOLS_TAG \
+		.build/var/REGISTRY \
+		$(shell find testing/deployer_helm_tiller_base/standard -type f) \
+    | .testing/marketplace/deployer/helm_tiller_onbuild
+	$(call print_target)
+	TEST_ID=$(TEST_ID) \
+	REGISTRY=$(REGISTRY) \
+	MARKETPLACE_TOOLS_TAG=$(MARKETPLACE_TOOLS_TAG) \
+	  ./testing/marketplace/deployer/helm_tiller_onbuild/standard/test
```
huyhg: `touch "$@"`

testing.Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+.PHONY: testing/marketplace/deployer/helm_tiller_onbuild
+testing/marketplace/deployer/helm_tiller_onbuild: \
```
huyhg: follow up: move `testing/py` target from the root `Makefile` here, and include `testing.Makefile` in the root `Makefile`.

testing.Makefile
```diff
@@ -0,0 +1,46 @@
+ifndef __TESTING_MAKEFILE__
+
+__TESTING_MAKEFILE__ := included
+
+include common.Makefile
+include gcloud.Makefile
```
huyhg: nit: this should be included at the top level instead, but let's make that change when we actually include this makefile. Can you add a TODO here?

testing/marketplace/deployer/helm_tiller_onbuild/helm-dependency-build/chart/chartmuseum-marketplace/requirements.yaml
```diff
@@ -0,0 +1,7 @@
+dependencies:
+- name: marketplace-integration
+  version: 0.0.1
```
huyhg: Just a note that this doesn't have a be a fixed version. This is where `requirements.yaml` and `requirements.yaml.lock` differ. The lock file points to a specific version, while the yaml file specifies the constraints of what versions are accepted. `helm dependency update` pulls the latest versions that satisfy the constraints, while `helm dependency build` downloads the tarballs of those specific versions.

See [here](https://github.com/Masterminds/semver#checking-version-constraints) for more info on how version constraints are specified.

Probably for this, we want to specify `0.0.x` versions?

testing/marketplace/deployer/helm_tiller_onbuild/helm-dependency-build/chart/chartmuseum-marketplace/requirements.yaml
```diff
@@ -0,0 +1,7 @@
...
+  repository: https://cloud-marketplace-tools-charts.storage.googleapis.com/
+- name: chartmuseum
+  version: 1.6.0
```
huyhg: Similarly, we want to specify `1.6.x` here.

testing/marketplace/deployer/helm_tiller_onbuild/helm-dependency-build/schema.yaml
```diff
@@ -0,0 +1,34 @@
...
+        roles:
+        - type: Role
+          rulesType: PREDEFINED
+          rulesFromRoleName: cluster-admin
+required:
+- name
+- namespace
+- application_uid
+- marketplace-integration.image
+- marketplace-integration.service_account
```
huyhg: I forgot--why does this need a service account again? I thought we're creating them directly in the hooks?

testing/marketplace/deployer/helm_tiller_onbuild/standard/Dockerfile
```diff
@@ -0,0 +1,3 @@
+ARG MARKETPLACE_TOOLS_TAG
```
huyhg: It's unfortunate that we have to have 2 separate similar sets of folders to test a rather small feature of the onbuild version. Can you follow up to clean this up? I think these should just be 2 test cases with 2 separate entry scripts.

testing/marketplace/deployer/helm_tiller_onbuild/helm-dependency-build/schema.yaml
trironkk: Agreed - done!

Ideally the make targets will mirror the file system - hopefully we can align those in the future.

testing/marketplace/deployer/helm_tiller_onbuild/standard/Dockerfile
trironkk: Certainly will follow up, though it's not clear to me what "clean" looks like. Let's discuss. :)

marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
trironkk: +1 - I was so sad when I had to add `true`.

`print_config.py` is pending an overhaul anyways - will resolve at that time.

testing/marketplace/deployer/helm_tiller_onbuild/helm-dependency-build/chart/chartmuseum-marketplace/requirements.yaml
trironkk: Done!

testing/marketplace/deployer/helm_tiller_onbuild/helm-dependency-build/chart/chartmuseum-marketplace/requirements.yaml
trironkk: Done!

trironkk:
Follow ups (appending other action items):
* [ ]  Support for original helm deployer
* [ ]  print_config.py support for extracting keys
* [ ]  Chart Release Process
* [ ]  Application Labels
* [ ]  Add entrypoint.sh to Dockerfile for deployers
* [ ]  Example repo refactor
* [ ]  DEPLOYER_IMAGE x-google-marketplace property for hooks images.
* [ ]  Block merging of PRs on integration testing (e.g. `testing/marketplace/deployer/helm_tiller_onbuild`)
* [ ]  Align make target names and directory structure.
* [ ]  Clean up testing paradigm within `testing/marketplace/deployer_helm_tiller_base/onbuild`.

PR #222, opened 2018-10-02
Correct bug in storage.py.

trironkk:


trironkk:
This was already fixed; I was looking at a stale view of origin master.

PR #223, opened 2018-10-04
Add validation checks for schema form field.

trironkk:


marketplace/deployer_util/config_helper.py
```diff
@@ -34,6 +34,8 @@
 XTYPE_STORAGE_CLASS = 'STORAGE_CLASS'
 XTYPE_STRING = 'STRING'
 
+FORM_TYPES = ('text',)
```
huyhg: use list: `['text']`

PR #224, opened 2018-10-05
Add APPLICATION_UID x-google-marketplace property.

trironkk:


marketplace/deployer_util/provision.py
```diff
@@ -80,6 +83,8 @@ def process(schema, values, deployer_image, deployer_entrypoint):
           schema, prop, app_name=app_name, namespace=namespace)
       props[prop.name] = value
       manifests += sc_manifests
+    elif prop.xtype == 'APPLICATION_UID':
```
huyhg: This is not the right place for it. We want to do this at configuration expansion time, similar to [this](https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/894cffecaa1cf9cb76cf7d5fb89a9dc52a3437f6/marketplace/deployer_util/expand_config.py#L69). The expansion is executed [here](https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/183d98a20cb2d49ca23d2181c665a189b5d03969/marketplace/deployer_util/deploy.sh#L40).

That way we shouldn't need to change anything in the backend.

trironkk: Discussed approach offline.

Done!

marketplace/deployer_util/config_helper.py
```diff
@@ -33,6 +33,7 @@
 XTYPE_SERVICE_ACCOUNT = 'SERVICE_ACCOUNT'
 XTYPE_STORAGE_CLASS = 'STORAGE_CLASS'
 XTYPE_STRING = 'STRING'
+XTYPE_APPLICATION_UID = 'APPLICATION_UID'
```
huyhg: Add validation that this field must be specified for helm-tiller deployer. Please follow up in another PR.

marketplace/deployer_util/expand_config.py
```diff
@@ -73,6 +77,10 @@ def expand(values_dict, schema):
       result[k] = generate_password(prop.password)
       continue
 
+    if v is None and prop.xtype == 'APPLICATION_UID' and app_uid:
```
huyhg: move `app_uid` check to an inner `if`, and should raise if it's not specified

trironkk: we can't actually do that, as expand_config.py needs to run before app_uid is specified.

i'll leave a comment explaining why.

trironkk: ... or maybe a bogus value is okay...? feels icky.

trironkk: Nevermind - `print_config.py` does not rely on `expand_config.py` having been run. Removed two-pass approach.

trironkk:
ready for review; please take a look!

marketplace/deployer_util/expand_config.py
```diff
@@ -73,6 +77,14 @@ def expand(values_dict, schema):
       result[k] = generate_password(prop.password)
       continue
 
+    if v is None and prop.xtype == 'APPLICATION_UID':
+      if not app_uid:
+        raise InvalidProperty(
+            'Property {} is of type APPLICATION_UID, but no value was specified'
```
huyhg: "no value" is misleading--change to "--app_uid was not specified".

There is a separate question whether we should respect a directly specified value. I think always overriding it is fine for now.

trironkk: Done!

PR #230, opened 2018-10-09
Limit the length of names of RBAC resources.

trironkk:
Also corrects the value passed to the labels.

### DIFF
--- a/marketplace/deployer_util/provision.py
+++ b/marketplace/deployer_util/provision.py
@@ -273,7 +273,7 @@ def provision_service_account(schema, prop, app_name, namespace):
         'kind': 'RoleBinding',
         'metadata': {
             'name':
-                limit_name('{}:{}:{}-rb'.format(app_name, prop.name, role)),
+                limit_name('{}:{}:{}-rb'.format(app_name, prop.name, role), 64),
             'namespace':
                 namespace,
         },
@@ -291,8 +291,9 @@ def provision_service_account(schema, prop, app_name, namespace):
         'kind': 'ClusterRoleBinding',
         'metadata': {
             'name':
-                limit_name('{}:{}:{}:{}-crb'.format(namespace, app_name,
-                                                    prop.name, role)),
+                limit_name(
+                    '{}:{}:{}:{}-crb'.format(namespace, app_name, prop.name,
+                                             role), 64),
             'namespace':
                 namespace,
         },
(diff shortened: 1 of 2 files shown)


marketplace/deployer_util/provision.py
```diff
@@ -272,8 +272,10 @@ def provision_service_account(schema, prop, app_name, namespace):
         'apiVersion': 'rbac.authorization.k8s.io/v1',
         'kind': 'RoleBinding',
         'metadata': {
-            'name': '{}:{}:{}-rb'.format(app_name, prop.name, role),
-            'namespace': namespace,
+            'name':
+                dns1123_name('{}:{}:{}-crb'.format(app_name, prop.name, role)),
```
huyhg: `dns1123_name` is not the right one to use for this, as this name doesn't need to conform to DNS spec. It also turns colons into dashes. I would suggest creating a new method to cap the length of the resources, and appending a hash suffix similar to `dns1123_name` method to avoid collisions.

Still think the colon convention is good. Here is an example of what `kubectl get clusterrolebindings` returns:

```
NAME                                                   AGE
gce:beta:kubelet-certificate-bootstrap                 83d
gce:beta:kubelet-certificate-rotation                  83d
system:aws-cloud-provider                              83d
system:basic-user                                      83d
system:controller:certificate-controller               83d
system:controller:ttl-controller                       83d
system:discovery                                       83d
```

huyhg: nit: s/crb/rb/

marketplace/deployer_util/provision.py
```diff
@@ -356,17 +359,19 @@ def dns1123_name(name):
   fixed = re.sub(r'[.]', '-', fixed)
   fixed = re.sub(r'[^a-z0-9-]', '', fixed)
   fixed = fixed.strip('-')
-  if len(fixed) > 64:
-    fixed = fixed[:59]
+  fixed = limit_name(fixed)
```
huyhg: According to the [RFC](https://tools.ietf.org/html/rfc1123.html#page-13), this can be up to 255. It's mandatory that it can be up to 63. I'll say stick with 63 (and correct my earlier typo of 64) here, but change the default of `limit_name` to something larger, like 127.

marketplace/deployer_util/provision.py
```diff
@@ -356,17 +359,19 @@ def dns1123_name(name):
...
+  return fixed
 
-  # Add a hash at the end if the name has been modified.
-  if fixed != name:
+def limit_name(name, length=64):
```
huyhg: can you add a separate test for this, in particular names with colons?

PR #235, opened 2018-10-10
Fix bug in print_config.py invocation.

trironkk:


marketplace/deployer_util/deploy.sh
```diff
@@ -37,8 +37,14 @@ handle_failure() {
 }
 trap "handle_failure" EXIT
 
-NAME="$(/bin/print_config.py --param '{"x-google-marketplace": {"type": "NAME"}}')"
-NAMESPACE="$(/bin/print_config.py --param '{"x-google-marketplace": {"type": "NAMESPACE"}}')"
+NAME="$(/bin/print_config.py \
```
huyhg: Can we add a flag to `print_config` to specify mode: PRE_EXPANDED or POST_EXPANDED (default)? It simplifies the flags here, avoids repetition, and helps with documenting what's going on here.

trironkk: Can I do this as a follow up, rather than in this PR?

I'd like to unbreak master.

huyhg: SG.

PR #239, opened 2018-10-12
Improve tools developing xp for Makefile users

huyhg:


huyhg:
Relevant example update: GoogleCloudPlatform/marketplace-k8s-app-example#108

app.Makefile
```diff
@@ -29,13 +30,82 @@ $(shell echo '$(APP_PARAMETERS)' '$(APP_TEST_PARAMETERS)' \
...
+
 .build/app: | .build
 	mkdir -p "$@"
 
 
+.build/app/base_deployer: | .build/app
+	mkdir -p "$@"
+
+
+##### Tools development targets #####
```
trironkk: this is aimed at making it convenient for users to leverage `marketplace-k8s-app-tools`'s build system, but I thought we wanted to reduce those dependencies.

ideally i think partners will not build any of our assets, and we're not far from that goal - what do you think?

huyhg: We would not encourage partners to use our Makefiles anymore. I am aware that several partners are already using them, so unless we have to, we don't want to kill them right now.

We still have dependencies on these Makefiles from example and c2d, and currently they don't work for local development. This is to address that broken workflow.

I don't mind deleting them all in the future. We need to figure out what we want to do at that time. I have a hunch that simply replacing Makefiles with bash scripts is not gonna be better.

trironkk: i worry that exposing adding local development targets to `app.Makefile` exposes our build implementation more broadly than necessary.

what do you think about adding this functionality to `marketplace.Makefile` instead?

scripts/derive_tag.sh
```diff
@@ -16,15 +16,71 @@
...
+function hashall() {
+  # Include the current checked-out commit in the hash.
+  git rev-parse --git-dir HEAD
+  while IFS= read -r line; do
+    # Remove the status column.
+    file=${line:3}
+    fullfile="$(git rev-parse --show-toplevel)/$file"
+    echo $fullfile
+    if [[ -f "$fullfile" ]]; then
+      git hash-object $fullfile
```
trironkk: can we quote `$fullfile`?

scripts/derive_tag.sh
```diff
@@ -16,15 +16,71 @@
...
+if [ ! -z "${FORCE_MARKETPLACE_TOOLS_TAG}" ]; then
+  >&2 echo "====== NOTICE ======"
+  >&2 echo "Using manual override of MARKETPLACE_TOOLS_TAG env var."
+  >&2 echo "===================="
+  echo -n "${FORCE_MARKETPLACE_TOOLS_TAG}"
+  exit 0
+fi
+
+
+if [[ "$(status)" == "" ]] && [[ "$(head)" == "" ]]; then
```
trironkk: is the second condition sufficient to determine that this isn't a git repo?

PR #241, opened 2018-10-13
Add --key to print_config.py.

trironkk:
Also refactor `--param` to `--xtype` and retrofitted existing invocations.


trironkk:
Totally agreed on your comment. Merging this to replace jq voodoo in https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/221, but will revisit later.

Issue #242, opened 2018-10-18
Helm tiller deployer base image


trironkk:
Partner Facing:
* [x]  DEPLOYER_IMAGE x-google-marketplace property for hooks images.
* [ ]  Example repo refactor
* [x] Discontinue pattern of using a wrapper chart - upstream `marketplace-integration` dependency instead.
* [x]  Chart Release Process

Cleanup:
* [ ]  print_config.py support for extracting keys
* [ ]  Application Labels
* [ ]  Add entrypoint.sh to Dockerfile for deployers
* [ ]  Block merging of PRs on integration testing (e.g. `testing/marketplace/deployer/helm_tiller_onbuild`)
* [ ]  Support for original helm deployer
* [x]  Align make target names and directory structure.
* [x]  Clean up testing paradigm within `testing/marketplace/deployer_helm_tiller_base/onbuild`.
* [x]  Investigate deploying to `kube-system` namespace.

eshiroma:
Closing as obsolete since Helm Tiller is deprecated, and #513 tracks deprecating this deployer. Though this issue may help guide some still-needed cleanup.

PR #243, opened 2018-10-18
Add DEPLOYER_IMAGE x-google-marketplace property.

trironkk:
Paired marketplace-k8s-app-example pull request (https://github.com/GoogleCloudPlatform/marketplace-k8s-app-example/pull/110).

This is a follow up task in https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/issues/242.


trironkk:
Will also update our backend to be consistent with `provision.py`.

marketplace/deployer_util/config_helper.py
```diff
@@ -201,6 +202,9 @@ def __init__(self, name, dictionary, required):
       elif xt == XTYPE_IMAGE:
         d = self._x.get('image', {})
         self._image = SchemaXImage(d)
+      elif xt == XTYPE_DEPLOYER_IMAGE:
+        d = self._x.get('image', {})
```
huyhg: s/image/deployerImage/

marketplace/deployer_util/provision.py
```diff
@@ -53,6 +53,16 @@ def process(schema, values, deployer_image, deployer_entrypoint):
   app_name = get_name(schema, values)
   namespace = get_namespace(schema, values)
 
+  # Inject DEPLOYER_IMAGE property values if not already present.
+  for key, value in values.items():
```
huyhg: I believe this has a bug.

We need to iterate over `schema.properties` and fine the `DEPLOYER_IMAGE` propertis. `values` wouldn't have that key. Then:

```
if key not in values:
  values[key] = deployer_image
```

marketplace/deployer_util/provision.py
```diff
@@ -53,6 +53,16 @@ def process(schema, values, deployer_image, deployer_entrypoint):
...
+    if value:
+      continue
+    if key not in schema.properties:
+      continue
+    if not schema.properties[key].xtype == 'DEPLOYER_IMAGE':
+      continue
+    values[key] = deployer_image
```
huyhg: can you extract this into a function and add a test case? This file is missing a lot of tests, but I think we can test this particular thing.

trironkk: Added a test, though I'm unsatisfied with it.

I'd be happy with a test that validated that `schema.yaml` + `values.yaml` -> `manifest.yaml`, but there's other higher-impact broader-scope tests that would add more value (e.g. sample deployers for each of the base deployers).

marketplace/deployer_util/config_helper_test.py
```diff
@@ -49,6 +49,11 @@
     default: gcr.io/google/busybox
     x-google-marketplace:
       type: IMAGE
+  propertyDeployerImage:
+    type: string
+    default: gcr.io/google/deployer
```
huyhg: rm default value

marketplace/deployer_util/provision.py
```diff
@@ -53,6 +53,9 @@ def process(schema, values, deployer_image, deployer_entrypoint):
   app_name = get_name(schema, values)
   namespace = get_namespace(schema, values)
 
+  # Inject DEPLOYER_IMAGE property values if not already present.
+  values = inject_deployer_image_properties(schema, deployer_image, values)
```
huyhg: optional:
- remove the return value to make the API cleaner: it's modifying the passed in values directly
- Make `values` the first argument

marketplace/deployer_util/provision_test.py
```diff
@@ -43,3 +46,17 @@ def test_limit_name(self):
...
+    properties:
+      deployer_image:
+        type: string
+        x-google-marketplace:
+          type: DEPLOYER_IMAGE
+    ''')
+    values = {}
+    deployer_image = 'gcr.io/cloud-marketplace/partner/solution/deployer:latest'
+    self.assertEquals(
+        provision.inject_deployer_image_properties(
```
huyhg: removing the return value will inconvenience the test--well I'm not opposed to having the method returning the input `values` for this purpose

PR #246, opened 2018-10-22
Remove 'Pod' filter from testing

jprzychodzen:


marketplace/deployer_util/run_tester.py
```diff
@@ -46,10 +46,6 @@ def main():
     full_name = "{}/{}".format(resource_def['kind'],
                                deep_get(resource_def, 'metadata', 'name'))
 
-    if resource_def['kind'] != 'Pod':
```
huyhg: This is necessary here since this loop checks for status of individual pods.

L35 is where the test manifests were applied. It doesn't filter out resources other than jobs. Did it not do what you expected in your testing?

PR #247, opened 2018-10-22
Convert integration tests to minimal helm charts.

trironkk:
Follow up from https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/245.


testing/marketplace/deployer_helm_tiller_base/onbuild/helm-dependency-build/chart/minimal-helm-chart/templates/pod.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Pod
+metadata:
+  name: {{ .Release.Name }}
+  labels:
+    heritage: {{ .Release.Service | quote }}
+    release: {{ .Release.Name | quote }}
+spec:
+  containers:
+  - name: {{ .Release.Name }}
+    image: gcr.io/cloud-marketplace-containers/google/nginx1:latest
```
huyhg: suggestion: use some simple thing like busybox with a sleep script instead of nginx. Just a note that if the script sleeps, make sure it can get interrupted or the pod won't shutdown.

trironkk: Done.

PR #249, opened 2018-10-26
Support custom timeout

jprzychodzen:


marketplace/deployer_util/run_tester.py
```diff
@@ -30,6 +30,7 @@ def main():
   parser = ArgumentParser(description=_PROG_HELP)
   parser.add_argument('--namespace')
   parser.add_argument('--manifest')
+  parser.add_argument('--timeout', type=int)
```
huyhg: should set a default value here so that the flag is not required

jprzychodzen: Actually, it was something I was considering, but... It's called from `deploy_with_tests.sh`, that defines it's own default. I do not feel that having multiple defaults would be good practice. Also, I think it's better to keep default value as top-level as possible, to gather them together.

Maybe then this should be required parameter? WDYT?

huyhg: `--timeout` is generally a flag which should have a reasonable default value. I don't feel strongly either way in this context, as `run_tester.py` is a private helper.

PR #252, opened 2018-11-01
Add integration tests to cloudbuild.yaml.

trironkk:
Also remove ubbagent submodule.

We should reorganize our Makefiles, too, but let's consider that out of scope for this pull request.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -35,7 +35,7 @@
 
 steps:
 
-- id: Build dev
+- id: &BuildDev Build dev
   name: gcr.io/cloud-builders/docker
   waitFor:
   - '-'
@@ -49,7 +49,7 @@ steps:
   - marketplace/dev/Dockerfile
   - .
 
-- id: Build deployer_envsubst
+- id: &BuildEnvsubst Build deployer_envsubst
   name: gcr.io/cloud-builders/docker
   waitFor:
   - '-'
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -123,6 +123,62 @@ steps:
...
+  - *CopyKubectlCredentials
+  env:
+  - 'KUBE_CONFIG=/workspace/.kube'
+  - 'GCLOUD_CONFIG=/workspace/.config/gcloud'
+  # Use local Docker network named cloudbuild as described here:
+  # https://cloud.google.com/cloud-build/docs/overview#build_configuration_and_build_steps
+  - 'EXTRA_DOCKER_PARAMS=--net cloudbuild'
+  args:
+  - make
+  - --file=testing.Makefile
```
huyhg: Can we include this file in the main `Makefile` instead?

testing.Makefile
```diff
@@ -73,4 +73,12 @@ testing/marketplace/deployer/helm_tiller_onbuild: \
 .PHONY: testing/marketplace/deployer/envsubst
 testing/marketplace/deployer/envsubst: \
 		.testing/marketplace/deployer/envsubst/standard
+
+
+.PHONY: testing/all
+testing/all: \
```
huyhg: Note that we currently also have target `testing/py` (which builds the py testing container) and `test/py` (which runs python tests).

I think the main Makefile can define `test/integrations` which should depend on the deployer targets.

trironkk:
Opened https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/issues/253 to track Makefile refactor.

PR #254, opened 2018-11-01
Add a cron job to garbage collect old test namespaces

huyhg:


huyhg:
We apply this only once for the cluster.

testing/apptest-gc/manifests.yaml
```diff
@@ -0,0 +1,84 @@
...
+      | jq -r \
+        '.items[]
+         | { "name": .metadata.name, "time": (now - (.metadata.creationTimestamp|fromdate)) }
+         | select(.name | test("^apptest-\\w{8,16}$"))
+         | select(.time >= 1800)
+         | .name'))
+
+    echo "Number of eligible namespaces: ${#namespaces[@]}"
+
+    DRYRUN=${DRYRUN:-true}
```
trironkk: Looks like DRYRUN defaults to true - does this mean that the Cron is disabled by default?

huyhg: Oops, meant to set an environment variable for the container spec.

testing/apptest-gc/manifests.yaml
```diff
@@ -0,0 +1,90 @@
...
+  namespace: apptest-namespaces-gc
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gc-script
+  namespace: apptest-namespaces-gc
+data:
+  gc.sh: |-
+    set -eo pipefail
```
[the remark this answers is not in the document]
huyhg: Done.

PR #264, opened 2018-11-14
Add marketplace-integration chart to CI/CD.

trironkk:
This pull request adds the marketplace-integration-testing chart to our CI/CD.

It does not add updating of `index.yaml` - that will be added to `cloudbuild-tag.yaml` in a follow up.


marketplace/charts/marketplace-integration.Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM gcr.io/cloud-builders/gcloud-slim
```
huyhg: I don't see this file used?

I remember there was a Dockerfile that is used for only side effects of publishing the chart to the GCS bucket. Is this the same one? We want to remove that too, right?

cloudbuild.yaml
```diff
@@ -170,3 +184,9 @@ images:
   - gcr.io/$PROJECT_ID/k8s/deployer_helm_tiller:sha_$SHORT_SHA
   - gcr.io/$PROJECT_ID/k8s/deployer_helm_tiller/onbuild:sha_$COMMIT_SHA
   - gcr.io/$PROJECT_ID/k8s/deployer_helm_tiller/onbuild:sha_$SHORT_SHA
+
+artifacts:
+  objects:
+    location: gs://$PROJECT_ID-charts/
```
huyhg: Shouldn't we use a bucket with the same name as the project and a path `/charts/`, i.e. `gs://$PROJECT_ID/charts/`?

marketplace/charts/marketplace-integration.Dockerfile
trironkk: You're right - this is that file. It was a mistake to include - removing.

cloudbuild.yaml
trironkk: I've created a different bucket in each project (i.e. `gs://$PROJECT_ID-charts/`). My thought process was that we would want to make the bucket public and such, so being able to manage iam processes for an isolated bucket would be easier.

What do you think?

huyhg: My thinking is that this project is intended for housing public artifacts, so I would just use the project as the bucket to house everything.

Just a note that this bucket is independent from the bucket used for GCR, so we can set different permissions there.

trironkk: Makes sense for `cloud-marketplace-tools` and `marketplace-k8s-app-tools`, but our pull requests are run within `cloud-marketplace-ops-test`, and some of the items in that project aren't intended to be public (e.g. reporting secrets).

trironkk: Updated `cloudbuild-tag.yaml` to publish to `gs://cloud-marketplace-tools/charts` as we discussed offline.

PR #265, opened 2018-11-14
Makefile refactoring.

trironkk:
Moved python testing to `testing.Makefile` and reformatted dependencies.


testing.Makefile
```diff
@@ -76,10 +72,44 @@ testing/marketplace/deployer/envsubst: \
 		.testing/marketplace/deployer/envsubst/standard
 
 
-.PHONY: testing/all
-testing/all: \
+.PHONY: testing/integration
```
huyhg: I prefer `test/integration` and `test/py`, or `tests/integration`. `testing` makes sense in this context since that's the name of the Makefile. Would recommend consider `tests`--and renaming accordingly.

testing.Makefile
```diff
@@ -76,10 +72,44 @@ testing/marketplace/deployer/envsubst: \
...
+PYTHON_TEST_DIRS = $(shell find . \
+                     -path ./vendor -prune -o \
+                     -name "*_test.py" -type f -print \
+                     | xargs -n 1 dirname | sort | uniq)
+
+# Append .__pytest__ to all python test directories to generate targets.
+PYTHON_TEST_TARGETS = $(foreach f,$(PYTHON_TEST_DIRS),$(f).__pytest__)
+
+
+.PHONY: test/py
```
huyhg: should be `testing/py` to match the following target

trironkk: Good eye. :)

testing.Makefile
trironkk: Renamed to `tests/*`.

PR #276, opened 2018-12-05
Only overwrite gcloud command if it exists

huyhg:
Application default credential config does not have
cmd-path set. The old code incorrectly set a new cmd-path,
which would make credential refreshing fail.


huyhg:
No idea why it failed the first time. Tiller said it timed out. Rerunning fixed it :-/

PR #282, opened 2018-12-18
Add deployer_helm_tiller to cloudbuild-tag.yaml publishing.

trironkk:


trironkk:
Would like to submit this before example repo refactor so that example repo can reference the 0.7.1 helm_tiller image.

PR #287, opened 2018-12-26
Add updating of index.yaml for marketplace-integration charts.

trironkk:


trironkk:
Will do - after merging, I'll reset the existing `index.yaml` for all of the projects and create `0.7.3`.

PR #289, opened 2019-01-07
[Testrunner] Fix Testrunner build

jprzychodzen:
Also:
* Update go rules
* Format with 'bazel build //:gazelle'


testrunner/runner/BUILD
```diff
@@ -12,14 +12,7 @@
...
-    "@io_bazel_rules_go//go:def.bzl",
-    "go_binary",
-)
-load(
-    "@io_bazel_rules_docker//go:image.bzl",
-    "go_image",
-)
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 
 binary_srcs = [
```
huyhg: rm these?

jprzychodzen: Removing 'load' line breaks the build. Is this some other suggestion?

Committing as is.

PR #304, opened 2019-01-29
Support boolean application create property

huyhg:
Some helm charts wish to use global.application.create value
to control whether the Application resource should be included
in the generated manifests. This is the negation of the truthy
value of the existing APPLICATION_UID property.

### DIFF
--- a/marketplace/deployer_helm_tiller_base/bin/deploy.sh
+++ b/marketplace/deployer_helm_tiller_base/bin/deploy.sh
@@ -48,12 +48,6 @@ NAMESPACE="$(/bin/print_config.py \
 export NAME
 export NAMESPACE
 
-app_uid=$(kubectl get "applications/$NAME" \
-  --namespace="$NAMESPACE" \
-  --output=jsonpath='{.metadata.uid}')
-
-/bin/expand_config.py --values_mode=raw --app_uid="$app_uid"
-
 bin/deploy_internal.sh
 
 patch_assembly_phase.sh --status="Success"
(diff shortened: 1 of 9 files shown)


tests/marketplace/deployer_helm_tiller_base/onbuild/standard/schema.yaml
```diff
@@ -12,6 +12,9 @@ properties:
     type: string
     x-google-marketplace:
       type: APPLICATION_UID
+      applicationUid:
+        generatedProperties:
+          createApplicationBoolean: global.application.create
```
trironkk: Can we change the `helm-dependency-build` variant of this test, too?

https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/tree/541156e4055c4940b205f732fbdf95c5c28b2dff/tests/marketplace/deployer_helm_tiller_base/onbuild/helm-dependency-build

marketplace/deployer_helm_tiller_base/bin/deploy.sh
```diff
@@ -48,12 +48,6 @@ NAMESPACE="$(/bin/print_config.py \
 export NAME
 export NAMESPACE
 
-app_uid=$(kubectl get "applications/$NAME" \
-  --namespace="$NAMESPACE" \
-  --output=jsonpath='{.metadata.uid}')
-
-/bin/expand_config.py --values_mode=raw --app_uid="$app_uid"
-
 bin/deploy_internal.sh
```
trironkk: would you mind making this fully qualified?

tests/marketplace/deployer_helm_tiller_base/onbuild/standard/schema.yaml
huyhg: Was thinking that we should use both variants in test. It's not a very systematic testing approaching (as the tests are intended to test different unrelated aspects), but it makes me feel better for now.

Let me know if you feel strongly. I can follow up.

PR #306, opened 2019-01-30
Delete testrunner now that it lives in its own repo

huyhg:


huyhg:
@jprzychodzen Testrunner is not really relevant to this repo. We will have pointers to it in other docs.

PR #307, opened 2019-02-01
Use variables in cloudbuild again

huyhg:


cloudbuild.yaml
```diff
@@ -151,7 +151,7 @@ steps:
   args:
   - -exc
   - |
-    gcloud container clusters get-credentials 'testing-marketplace' --region 'us-central1' --project '$PROJECT_ID'
+    gcloud container clusters get-credentials '${_CLUSTER_NAME}' --zone '${_CLUSTER_LOCATION}' --project '$PROJECT_ID'
```
jprzychodzen: What do you think about adding sane defaults, for reference purpose, if one would fork this repo?

Something like [this](https://github.com/GoogleCloudPlatform/click-to-deploy/blob/master/cloudbuild.yaml#L22,L24)?

PR #315, opened 2019-02-15
Improve Ingress readiness

wgrzelak:
Checking Ingress's readiness is more difficult. Even if the IP address is available, it doesn't mean that the ingress is ready. We should wait for annotations instead.

Not ready ingress:

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"extensions/v1beta1","kind":"Ingress","
...
```

Ready ingress:

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    ingress.kubernetes.io/backends: '{"k8s-be-32534--85c24f04c12b3680":"HEALTHY"}'
    ingress.kubernetes.io/forwarding-rule: k8s-fw-wgrzelak-wp-22768-wordpress-ingress--85c24f04c12b3680
    ingress.kubernetes.io/https-forwarding-rule: k8s-fws-wgrzelak-wp-22768-wordpress-ingress--85c24f04c12b3680
...
```

### DIFF
--- a/marketplace/deployer_util/wait_for_ready.py
+++ b/marketplace/deployer_util/wait_for_ready.py
@@ -16,6 +16,7 @@
 
 import sys
 import time
+import json
 
 from argparse import ArgumentParser
 from bash_util import Command
@@ -178,8 +179,12 @@ def is_service_ready(resource):
 
 
 def is_ingress_ready(resource):
-  if 'ingress' in resource['status']['loadBalancer']:
-    return True
+  for key, value in annotations(resource).items():
+    if key == 'ingress.kubernetes.io/backends':
+      data = json.loads(value)
+      status = list(data.values())[0]
+      if status == "HEALTHY":
+        return True
 
   log("INFO Ingress/{} is not ready.".format(name(resource)))
   return False
@@ -198,6 +203,10 @@ def name(resource):
 
 def labels(resource):
   return resource['metadata']['labels']
+
+
+def annotations(resource):
+  return resource['metadata']['annotations']
 
 
 def total_replicas(resource):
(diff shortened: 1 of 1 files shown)


marketplace/deployer_util/wait_for_ready.py
```diff
@@ -200,6 +205,10 @@ def labels(resource):
   return resource['metadata']['labels']
 
 
+def annotations(resource):
+  return resource['metadata']['annotations']
```
huyhg: use `get('metadata', {}).get('annotations', {})` to be safe? I know we should do that everywhere too, but at least this is an improvement. We can eventually change the other places as well.

marketplace/deployer_util/wait_for_ready.py
```diff
@@ -178,8 +179,12 @@ def is_service_ready(resource):
 
 
 def is_ingress_ready(resource):
-  if 'ingress' in resource['status']['loadBalancer']:
-    return True
+  for key, value in annotations(resource).items():
+    if key == 'ingress.kubernetes.io/backends':
```
huyhg: I believe these annotations are gce-ingress specific; i.e. it only works for GKE. I don't know of a universal way to determine if an ingress is ready. My suggestions:
- If this is critical to the applications that we're testing, we can keep this logic and add a TODO(issue) to follow up.
- Otherwise, skip ingresses when checking for readiness.

wgrzelak: We have workaround for Ingress, see https://github.com/GoogleCloudPlatform/click-to-deploy/blob/master/k8s/wordpress/apptest/tester/tester.sh#L20-L30. However, having this integrated with the framework sounds better to me.

Do you think we should skip Ingress anyway?

huyhg: It's reasonable to do this. However, we should separate two cases:
- `ingress.kubernetes.io/backends` annotation exists
  - Return true iff at least one backend is `HEALTHY`
- `ingress.kubernetes.io/backends` annotation does not exist
    - We can fallback to the previous (imperfect) logic

PR #318, opened 2019-02-22
Update schema to support version 2

huyhg:


marketplace/deployer_util/config_helper.py
```diff
@@ -97,6 +97,11 @@ def load_yaml(yaml_str):
     return Schema(yaml.load(yaml_str))
 
   def __init__(self, dictionary):
+    self._x_google_marketplace = None
```
thiagolacerda:
```suggestion
    self._x_google_marketplace = dictionary.get('x-google-marketplace', None)
```

marketplace/deployer_util/config_helper.py
```diff
@@ -97,6 +97,11 @@ def load_yaml(yaml_str):
...
+    if 'x-google-marketplace' in dictionary:
```
thiagolacerda: with the suggested change above we can remove this if statement and its content

marketplace/deployer_util/config_helper.py
```diff
@@ -119,7 +124,12 @@ def __init__(self, dictionary):
 
   def validate(self):
     """Fully validates the schema, raising InvalidSchema if fails."""
-    if self.app_api_version is None:
+    is_v2 = False
+    if self._x_google_marketplace is not None:
+      self._x_google_marketplace.validate()
```
thiagolacerda: would this function raise an exception or something? feels weird to have this function call alone with no return statement, no parameters, no try catch... seems like we don't know what to expect from it.
Maybe include validate in the is_v2 function code block?

marketplace/deployer_util/config_helper.py
```diff
@@ -156,8 +171,268 @@ def properties_matching(self, definition):
...
+    self._published_version_meta = None
+    self._images = None
+    self._cluster_constraints = None
+
+    self._schema_version = dictionary.get('schemaVersion', _SCHEMA_VERSION_1)
+    if self._schema_version not in _SCHEMA_VERSIONS:
+      raise InvalidSchema('Invalid schema version {}'.format(
+          self._schema_version))
+
+    if 'clusterConstraints' in dictionary:
```
thiagolacerda: why don't use the same get with default option as line 189 above?

marketplace/deployer_util/config_helper.py
```diff
@@ -156,8 +171,268 @@ def properties_matching(self, definition):
...
+      self._cluster_constraints = SchemaClusterConstraints(
+          dictionary['clusterConstraints'])
+
+    if not self.is_v2():
+      return
+
+    if 'applicationApiVersion' not in dictionary:
```
thiagolacerda: from this line to line 217: Are those checks to validate that this schema is a v2? If so, we could put all that in a private validation function to remove all that code from this constructor. what do you think?

marketplace/deployer_util/config_helper.py
```diff
@@ -156,8 +171,268 @@ def properties_matching(self, definition):
...
+
+
+class SchemaClusterConstraints:
+  """Accesses top level clusterConstraints."""
+
+  def __init__(self, dictionary):
+    self._k8s_version = dictionary.get('k8sVersion', None)
+    self._resources = None
+
+    if 'resources' in dictionary:
```
thiagolacerda: use get with default to None?

marketplace/deployer_util/config_helper.py
```diff
@@ -156,8 +171,268 @@ def properties_matching(self, definition):
...
+
+class SchemaResourceConstraints:
+  """Accesses a single resource's constraints."""
+
+  def __init__(self, dictionary):
+    self._replicas = dictionary.get('replicas', None)
+    self._affinity = None
+    self._requests = None
+
+    if 'affinity' in dictionary:
```
thiagolacerda: ditto about get function (line 282 too)

marketplace/deployer_util/config_helper_test.py
```diff
@@ -514,6 +520,116 @@ def test_unknown_type(self):
                   type: UNKNOWN
             """))
 
+  def test_v2_fields(self):
```
thiagolacerda: nit: there are multiple asserts here, I'm usually a fan of 1:1 mapping of test case and asserts. My suggestion would be to have the load_yaml call in a setup function and have one test case per assert

huyhg: All of these fields are required in v2, so we have to specify all of them anyway.

marketplace/deployer_util/config_helper.py
```diff
@@ -441,7 +694,26 @@ def base64_encoded(self):
...
 
 class SchemaXReportingSecret:
-  """Wrapper class providing convenient access to REPORTING_SECRET property."""
+  """Accesses REPORTING_SECRET property."""
 
   def __init__(self, dictionary):
     pass
+
+
+def _get_or_invalid(dictionary, key, error_msg):
```
thiagolacerda: nit: when I saw the invocations of this function I could not infer what it was doing... can we rename to something more descriptive? like: _get_or_throw[_error]

marketplace/deployer_util/config_helper.py
```diff
@@ -441,7 +694,26 @@ def base64_encoded(self):
...
+  if key not in dictionary:
+    raise InvalidSchema(error_msg)
+  return dictionary[key]
+
+
+def _apply_or_none(dictionary, key, apply_fn):
```
thiagolacerda: ditto: could be something like: apply_if_exists or apply_if_not_none

marketplace/deployer_util/config_helper.py
```diff
@@ -441,7 +694,26 @@ def base64_encoded(self):
...
+  """Returns the result of apply_fn on the value if not None."""
+  if key not in dictionary:
+    return None
+  return apply_fn(dictionary[key])
+
+
+def _apply_or_invalid(dictionary, key, apply_fn, error_msg):
```
thiagolacerda: ditto about naming

marketplace/deployer_util/config_helper.py
huyhg: the original name was `_get_or_raise_invalid_schema`, but it was too long and I reduced it to this. I changed this to `_must_get`.

marketplace/deployer_util/config_helper.py
huyhg: changed to `_maybe_get_and_apply`

marketplace/deployer_util/config_helper.py
huyhg: changed to `_must_get_and_apply`.

PR #320, opened 2019-03-05
 Adapting deployer to work with istio

vcanaa:
Workaround for this issue in deployer:
https://github.com/istio/istio/issues/12187


scripts/verify
```diff
@@ -31,6 +31,10 @@ case $i in
     wait_timeout="${i#*=}"
     shift
     ;;
+  --istio=*)
```
huyhg: `instio_injection`?

marketplace/deployer_util/provision.py
```diff
@@ -208,6 +208,11 @@ def provision_deployer(schema, app_name, namespace, deployer_image,
           },
           'spec': {
               'template': {
+                  'metadata': {
```
huyhg: Please add a comment here, linking to external issues.

PR #322, opened 2019-03-07
Use full app CRD resource name

### LINKED ISSUE #321: Use fully qualified CRD name in deploy scripts
Deploy scripts that reference the Applications CRD (or others) should use the full resource name i.e. `applications.app.k8s.io` instead of `applications` .

The latter conflicts with any resource named `applications` in a different API group e.g. `applications.<customer api group>` and will break any marketplace deploys in that environment.

https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/master/scripts/watch.sh#L46 is an example which should be hardened.

vkamra:
Use `applications.app.k8s.io` instead of the short name for the applications CRD.

This hardens the tooling for environments where there might be a conflict on 
the `applications` short name and some other API group is prioritized by Kubernetes
discovery.

Fixes #321

### DIFF
--- a/marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
+++ b/marketplace/deployer_helm_tiller_base/bin/deploy_internal.sh
@@ -16,10 +16,10 @@
 
 set -eox pipefail
 
-app_uid=$(kubectl get "applications/$NAME" \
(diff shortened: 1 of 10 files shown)


vkamra:
@huyhg - Please take a look.

huyhg:
@vkamra You're awesome. Thank you!

PR #323, opened 2019-03-09
 Adding istio types in schema files

vcanaa:
Introducing schema file x-google-marketplace properties:
- Compatibility flag
```yaml
x-google-marketplace:
  istio:
    isSupported: True # Indicates whether the app is compatible with Istio.
```
- Deployer property
```yaml
properties:
- name: istioEnabled
  type: ISTIO_ENABLED # Indicates whether the app is being deployed with Istio.
```

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -363,6 +363,7 @@ properties:
             resources: ['EtcdCluster']
             verbs: ['*']
 ```
+
 ---
 
 ### type: STORAGE_CLASS
@@ -380,6 +381,8 @@ properties:
       storageClass:
         type: SSD
 ```
+
+---
 
 ### type: STRING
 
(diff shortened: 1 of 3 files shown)


docs/schema.md
```diff
@@ -400,6 +403,41 @@ properties:
...
+- `True`: The app is compatible with Istio.
+- `False`: The app is NOT compatible with Istio.
+
+```yaml
+properties:
+  # Property definitions...
+required:
+  # Required properties...
+x-google-marketplace:
+  istioCompatibility: True # or False
```
huyhg: I predict that there will be additional parameters here. Can we make this into a struct:

```
x-google-marketplace:
  istio:
    isSupported: True
    # Future...
    version: '>0.5'
```

vcanaa: That is a very good idea.

marketplace/deployer_util/config_helper.py
```diff
@@ -334,6 +343,20 @@ def memory(self):
     return self._memory
 
 
+class SchemaIstio:
+  """Accesses top level istio."""
+
+  def __init__(self, dictionary):
+    self._is_supported = dictionary.get('isSupported', None)
+
+    if 'isSupported' in dictionary:
```
huyhg: rm?

PR #326, opened 2019-03-13
Set ISTIO_ENABLED=true in verify for istio=enabled

vcanaa:
The verify script needs to set the ISTIO_ENABLED to true to communicate the deployer that Istio should be handled.

### DIFF
--- a/scripts/verify
+++ b/scripts/verify
@@ -91,9 +91,10 @@ if [[ "$istio" = "enabled" ]]; then
   echo "INFO Enabling istio in \"$NAMESPACE\" (istio-injection=enabled)."
   kubectl label namespace "$NAMESPACE" istio-injection=enabled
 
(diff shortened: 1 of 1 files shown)


scripts/verify
```diff
@@ -90,6 +90,18 @@ kubectl create namespace "$NAMESPACE"
 if [[ "$istio" = "enabled" ]]; then
   echo "INFO Enabling istio in \"$NAMESPACE\" (istio-injection=enabled)."
   kubectl label namespace "$NAMESPACE" istio-injection=enabled
+
+  istio_enabled_key="$(extract_schema_key.py \
```
huyhg: it's possible that there are multiple ISTIO_ENABLED keys. Use `head -n 1` to make sure we only get the first key

Can we add a TODO and open an issue to track handling of multiple keys?

PR #329, opened 2019-03-14
Support INGRESS_AVAILABLE and add tests

huyhg:


docs/schema.md
```diff
@@ -410,7 +411,11 @@ In the example above, manifests can reference to the password as `explicitPasswo
 
 This boolean property receives a True value if the environment is detected to have Istio enabled, and False otherwise. The deployer and template can take this signal to adapt the deployment accordingly.
 
-If `ISTIO_ENABLED` is not present, it is assumed Istio is NOT enabled for the deployment.
```
vcanaa: Did you intentionally remove this?

huyhg: Yea, because it's not quite accurate. The application can decide to not have such a property if its manifests work in both cases, for example.

PR #331, opened 2019-03-16
Analyse cluster to set ISTIO_ENABLED flag accordingly

vcanaa:
This detects Istio installation and sets the ISTIO_ENABLED flag that is going to be used  by the manifest to provide extra k8s resources to work with Istio.

This still requires validation and unittests. This is an early PR to collect some feedback.

### DIFF
--- a/marketplace/deployer_util/provision.py
+++ b/marketplace/deployer_util/provision.py
@@ -17,6 +17,7 @@
 import hashlib
 import re
 from argparse import ArgumentParser
+from dict_util import deep_get
 
 import yaml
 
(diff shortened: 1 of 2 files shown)


marketplace/deployer_util/provision.py
```diff
@@ -104,6 +105,92 @@ def process(schema, values, deployer_image, deployer_entrypoint):
...
+        print_call=True).output
+    if policy == "policy: enabled":
+      return True
+    else:
+      return False
+  except:
+    print("istio-sidecar-injector policy.")
+    return False
+
+  return True
```
huyhg: this should never run

PR #335, opened 2019-03-27
Prefer ownerReference boolean over patching YAML

kellinm:
- According to https://github.com/kubernetes-sigs/application
  if spec.addOwnerRef is set to true then the owner reference
  will be inserted to matched objects. Use this behavior instead
  of using a yaml patch.

Signed-off-by: Robert Marshall <rmarshall@gitlab.com>


huyhg:
Thanks for the PR! It is our intention to adopt this soon.

However, this feature of the App controller is fairly new and isn't yet available in production environment. So in the mean time, we still need to do the patching.

I will leave this PR open and resolve it once we're ready.

Thanks again!

kellinm:
This looks like it's not applicable anymore as `ownerReference` doesn't even exist in the codebase.

Closing.

PR #337, opened 2019-03-31
Adding TLS_CERTIFICATE property.

wgrzelak:
# type: TLS_CERTIFICATE

This property provides an SSL/TLS certificate for the Kubernetes manifest. By default, a self-signed certificate is generated.

Example:

```yaml
properties:
  certificate:
    type: string
    x-google-marketplace:
      type: TLS_CERTIFICATE
      tlsCertificate:
        generatedProperties:
          base64EncodedPrivateKey: TLS_CERTIFICATE_KEY
          base64EncodedCertificate: TLS_CERTIFICATE_CRT
```

* `base64EncodedPrivateKey`: Denotes the name of a property receive a private key.
* `base64EncodedCertificate`: Denotes the name of a property receive a certificate.

You can provide your custom certificate by overwrite the `certificate` property in the JSON format as following:

```json
{
  "private_key": "-----BEGIN PRIVATE KEY-----\n...\n-----END PRIVATE KEY-----",
  "certificate": "-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----"
}
```

In your helm chart, based on the above example, you can handle the certificate:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: tls-secret
  namespace: demo
data:
  tls.key: {{ .Values.TLS_CERTIFICATE_KEY }}
  tls.crt: {{ .Values.TLS_CERTIFICATE_CRT }}
type: kubernetes.io/tls
```

In your envsubst manifest, based on the above example, you can handle the certificate:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: tls-secret
  namespace: demo
data:
  tls.key: $TLS_CERTIFICATE_KEY
  tls.crt: $TLS_CERTIFICATE_CRT
type: kubernetes.io/tls
```


docs/schema.md
```diff
@@ -262,6 +264,7 @@ It defines how this object will be handled. Each type has a different set of pro
 - `APPLICATION_UID`: The uuid of the created `Application` object.
 - `ISTIO_ENABLED`: Indicates whether Istio is enabled for the deployment.
 - `INGRESS_AVAILABLE`: Indicates whether the cluster is detected to have Ingress support.
+- `CERTIFICATE`: To be used to support a custom certificate or generate a self-signed certificate.
```
huyhg: TLS_CERTIFICATE

wgrzelak: Done, but I'm not sure about field under `type: TLS_CERTIFICATE`. Should it be (1) `certificate` or (2) `tlsCertificate`?

1)
```yaml
properties:
  certificate:
    type: string
    x-google-marketplace:
      type: TLS_CERTIFICATE
      certificate:
        generatedProperties:
          base64EncodedPrivateKey: TLS_CERTIFICATE_KEY
          base64EncodedCertificate: TLS_CERTIFICATE_CRT
```

2)
```yaml
properties:
  certificate:
    type: string
    x-google-marketplace:
      type: TLS_CERTIFICATE
      tlsCertificate:
        generatedProperties:
          base64EncodedPrivateKey: TLS_CERTIFICATE_KEY
          base64EncodedCertificate: TLS_CERTIFICATE_CRT
```

I think the option 2 is more clear, however it doesn't follow current convention. Please let me know, what do you think.

wgrzelak: I changed to `tlsCertificate`. PTAL

wgrzelak:
@huyhg @eshiroma @vcanaa gentle ping on this.

docs/schema.md
```diff
@@ -419,6 +422,66 @@ This boolean property receives a True value if the cluster is detected to have I
...
+metadata:
+  name: tls-secret
+  namespace: demo
+data:
+  tls.key: {{ .Values.TLS_CERTIFICATE_KEY }}
+  tls.crt: {{ .Values.TLS_CERTIFICATE_CRT }}
+type: kubernetes.io/tls
+```
+
+In your envsubst manifest, based on the above example, you can handle the certificate:
```
huyhg:
```
`envsubst`
```

marketplace/deployer_util/expand_config.py
```diff
@@ -70,12 +73,16 @@ def expand(values_dict, schema, app_uid=''):
   for k, prop in schema.properties.iteritems():
     v = values_dict.get(k, None)
 
+    # The value is not delivered to the framework, so it can be filled by the framework.
```
huyhg: Thanks for adding the comment! I would suggest the following wording:

```
# The value is not explicitly specified and thus is eligible for auto-generation.
```

marketplace/deployer_util/expand_config.py
```diff
@@ -84,18 +91,28 @@ def expand(values_dict, schema, app_uid=''):
         v = True
       elif prop.default is not None:
         v = prop.default
-    else:  # if v is None
+
+    # The value is not empty, so it can be expanded to the special properties.
+    # For example, property IMAGE can be expanded from the raw string image,
+    # to the generatedProperties properties.
+    if v is not None:
```
huyhg: Would suggest:

```
# Generate additional properties from this property.
```

marketplace/deployer_util/expand_config.py
```diff
@@ -84,18 +91,28 @@ def expand(values_dict, schema, app_uid=''):
...
           raise InvalidProperty(
               'Invalid value for STRING property {}: {}'.format(k, v))
         generate_properties_for_string(prop, v, generated)
+      elif prop.tls_certificate:
+        if not isinstance(v, str):
+          raise InvalidProperty(
+              'Invalid value for TLS_CERTIFICATE property {}: {}'.format(k, v))
+        generate_properties_for_tls_certificate(prop, v, generated)
 
+    # At this point, the property is populated and expanded, so overwrite the returned value.
```
huyhg: This is not overriding--this is recording the value of the property. I think we need to clarify what `result` and `generated` dictionaries capture. Can you update L71-73 and L122 with the following:

```python
### L71-73 ###

# Captures the final property name-value mappings.
# This has both properties directly specified under schema's `properties` and
# generated properties. See below for details about generated properties.
result = {}
# Captures only the generated properties. These are not directly specified in the schema
# under `properties`. Rather, their name are specified in special `generatedProperties` fields
# under individual property `x-google-marketplace`.
# Note that properties with generated values are NOT generated properties. 
generated = {}

# Copy explicitly specified values and generate values into result.

### L122 ###
# Copy generated properties into result, validating that there are no collisions.
```

marketplace/deployer_util/expand_config_test.py
```diff
@@ -106,6 +111,103 @@ def test_generate_properties_for_string_base64_encoded(self):
...
+        """)
+    result = expand_config.expand({}, schema)
+    cert = json.loads(result['c1'])
+    self.assertIsNotNone(result['c1'])
+    self.assertEqual(result['c1.Base64Key'],
+                     base64.b64encode(cert['private_key']))
+    self.assertEqual(result['c1.Base64Crt'],
+                     base64.b64encode(cert['certificate']))
+
+  def test_generate_certificate_verify(self):
```
huyhg: combine this with the previous one

tests/marketplace/deployer_helm_tiller_base/onbuild/standard/chart/minimal-helm-chart/values.yaml
```diff
@@ -0,0 +1,3 @@
+tls:
```
huyhg: what's the reason for expliciting specifying nulls here?

marketplace/deployer_util/expand_config_test.py
wgrzelak: Done.

marketplace/deployer_util/expand_config.py
wgrzelak: Done.

marketplace/deployer_util/expand_config.py
wgrzelak: Done.

marketplace/deployer_util/expand_config.py
wgrzelak: Done.

docs/schema.md
wgrzelak: Done.

tests/marketplace/deployer_helm_tiller_base/onbuild/standard/chart/minimal-helm-chart/values.yaml
wgrzelak: Having the values into the `values.yaml` file is not required, however it seems to be a good practice to maintain a full working helm chart, so the end-user can use `--set` flags to provide the certificate during the manifest is expanding.

PR #338, opened 2019-04-03
Reference schema.yaml in 'Building Application Deployer' doc

eshiroma:


docs/building-deployer.md
```diff
@@ -36,6 +36,9 @@ Decide how to craft your application k8s manifests.
   k8s. The downside is that templating capabilities are very limited.
   Follow instructions [here](building-deployer-envsubst.md).
 
+Include `schema.yaml`, a declaration of parameters required for provisioning the
+application. Follow instructions [here](schema.md) to craft this file.
```
huyhg: `schema.md` was intended to serve as a reference. The individual guides above should provide more context around how the file should be crafted and used.

Maybe this should be worded as:

```markdown
Note that whichever deployer guide you choose to follow, a `schema.yaml` file is required. It declares parameters required for provisioning the application. See [here](schema.md) for a reference of this file.
```

eshiroma: Done.

Issue #344, opened 2019-04-25
Remove flag that adds additional role binding to deployer


vcanaa:
Some deployers running in test mode need cluster wide access, which is more powerful than the official namespace access. In order to unblock verification for these particular apps, a flag for adding additional role bindings was introduced.

This flag can be removed once verification does not require cluster wide access for these apps.

eshiroma:
Verification is expected to leverage the same execution model.

PR #345, opened 2019-04-26
Improving verification logs + allowing deployer to use cluster-admin privileges in verification.

vcanaa:
Merge ruela/verification first

Allowing deployer to test some apps as cluster-admin, since it needs to poll some cluster wide resources to detect that the app is ready.

#344


marketplace/deployer_envsubst_base/create_manifests.sh
```diff
@@ -45,6 +45,8 @@ mkdir "$manifest_dir"
 if [[ "$mode" = "test" ]]; then
   if [[ -e "/data-test" ]]; then
     cp -RT "/data-test" "/data"
+  else
+    echo "$LOG_SMOKE_TEST ERROR Missing /data-test."
```
huyhg: Is this an error? We'll simply skip if there is no `/data-test`, right?

marketplace/deployer_util/run_tester.py
```diff
@@ -73,30 +79,31 @@ def main():
...
-          full_name, namespace)).output)
-
-
-def log(msg):
-  sys.stdout.write(msg + "\n")
-  sys.stdout.flush()
+def print_tester_logs(full_name, namespace):
+  try:
+    Command(
+        '''kubectl logs {} --namespace="{}"'''.format(full_name, namespace),
```
huyhg: single quotes instead?

marketplace/deployer_util/log_util.py
```diff
@@ -0,0 +1,23 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+
+def log(msg, *args):
```
huyhg: Can we mimic glog so that in the future we can use it as a drop-in replacement for this? This gets rid of the `ERROR` and `INFO` strings in individual log messages.

```
log.info()
log.warn()
log.error()
log.fatal()
```

(We shouldn't use glog now as its default configuration doesn't meet our need, which would require every binary to be changed to configure glog before it starts).

marketplace/deployer_util/overlay_test_schema.py
```diff
@@ -41,6 +44,12 @@ def main():
   dest = load_yaml(args.dest)
   if 'properties' in dest:
     for prop in orig['properties']:
+      if (deep_get(orig, 'properties', prop, 'x-google-marketplace', 'type') !=
+          deep_get(dest, 'properties', prop, 'x-google-marketplace', 'type')):
+        log(
+            "{} WARNING Changing x-google-marketplace type is not allowed. Property: {}",
```
huyhg: shouldn't this be an fatal error instead of warning?

marketplace/deployer_util/constants.py
```diff
@@ -13,3 +13,5 @@
 # limitations under the License.
 
 GOOGLE_CLOUD_TEST = 'marketplace.cloud.google.com/verification'
+
+LOG_SMOKE_TEST = 'LOG_SMOKE_TEST'
```
huyhg: I wonder if this is providing good enough value to justify over the inconsistency with what we currently have in our code. It seems that it's clear enough that these messages are talking about the tests.

My recommendation is to remove these log prefixes, in both sh and py.

scripts/verify
```diff
@@ -35,6 +35,10 @@ case $i in
     istio="${i#*=}"
     shift
     ;;
+  --role=*)
```
huyhg: Not clear what this is for. `--additional_clusterrole`?

scripts/verify
```diff
@@ -124,9 +128,19 @@ function clean_and_exit() {
   [[ -z "$message" ]] || echo "$message"
 
   delete_namespace
+  if [[ ! -z "$deployer_role_name" ]]; then
+    kubectl delete clusterrolebinding "$deployer_role_name"
+  fi
   exit 1
 }
 
+# TODO(#344) Remove
```
huyhg: - Move this to the line where the flag is defined.
- Clarify: `TODO(#344) Remove this flag once we have better support`

marketplace/deployer_envsubst_base/create_manifests.sh
vcanaa: It is an error. But the test pipeline should keep going to collect more results to show in the report.

marketplace/deployer_util/constants.py
vcanaa: I am adding them to help parse out for the report. Think of them as something that will be consumed by our code down the road. It will help have them label so we can control what to show and where.

marketplace/deployer_util/overlay_test_schema.py
vcanaa: I was originally inclined to treat as a warning and let the test itself make the decision. But maybe it is better to fail fast here.

scripts/verify
vcanaa: What about `--additional_deployer_role`

marketplace/deployer_envsubst_base/create_manifests.sh
huyhg: this variable is not defined in this file

marketplace/deployer_helm_base/create_manifests.sh
```diff
@@ -67,6 +67,11 @@ extract_manifest "$data_dir"
 if [[ "$mode" = "test" ]]; then
   extract_manifest "$test_data_dir"
 
+  if [[ ! -e "$data_dir/extracted" ]]; then
+    echo "$LOG_SMOKE_TEST No test charts declared."
```
huyhg: this variable is not defined in this file

marketplace/deployer_util/constants.py
huyhg: Maybe we should drop the `LOG_` prefix

marketplace/deployer_util/wait_for_ready.py
```diff
@@ -30,8 +30,9 @@ def main():
   parser.add_argument('--timeout')
   args = parser.parse_args()
 
-  log("INFO Wait {} seconds for the application '{}' to get into ready state"
-      .format(args.timeout, args.name))
+  log.info(
+      "Wait {} seconds for the application '{}' to get into ready state".format(
```
huyhg: rm `format`

marketplace/deployer_util/wait_for_ready.py
```diff
@@ -65,20 +66,22 @@ def main():
...
-      log("INFO Initialization: Found applications.app.k8s.io/{} ready status to be {}."
+      log.info(
+          "Initialization: Found applications.app.k8s.io/{} ready status to be {}."
           .format(args.name, healthy))
       previous_healthy = healthy
       if healthy:
-        log("INFO Wait {} seconds to make sure app stays in healthy state."
-            .format(min_time_before_healthy))
+        log.info(
+            "Wait {} seconds to make sure app stays in healthy state.".format(
```
huyhg: rm `format`

marketplace/deployer_util/wait_for_ready.py
```diff
@@ -65,20 +66,22 @@ def main():
...
     healthy = True
     for resource in top_level_resources:
       healthy = is_healthy(resource)
       if not healthy:
         break
 
     if previous_healthy != healthy:
-      log("INFO Initialization: Found applications.app.k8s.io/{} ready status to be {}."
+      log.info(
+          "Initialization: Found applications.app.k8s.io/{} ready status to be {}."
```
huyhg: rm `format`

PR #351, opened 2019-05-15
Handle DEPLOYER_IMAGE in v2 deployer

huyhg:


marketplace/deployer_util/expand_config.py
```diff
@@ -163,6 +165,17 @@ def validate_value_types(values, schema):
               k, prop.type, v))
 
 
+def maybe_derive_deployer_image(schema, values_dict):
```
vcanaa: Is this `maybe_...` a convention somewhere else?

Could you use this instead:
`derive_deployer_image_if_applicable`

huyhg: Resolved offline.

PR #354, opened 2019-05-21
Add 'enum' property to doc with sample

jprzychodzen:


docs/schema.md
```diff
@@ -223,6 +223,23 @@ Explanation of what the property is or what is used for. Be mindful of good expl
 
 If user does not provide a value, `default` will be used.
 
+### enum
+
+Array property, that will create combo box with a list of provided values, instead of free form input field.
+
+```yaml
+  some.property:
+    type: string
```
wgrzelak: Maybe it's worth to add that other types are also supported.

docs/schema.md
```diff
@@ -223,6 +223,23 @@ Explanation of what the property is or what is used for. Be mindful of good expl
 
 If user does not provide a value, `default` will be used.
 
+### enum
+
+Array property, that will create combo box with a list of provided values, instead of free form input field.
```
huyhg: Would suggest the following wording:

```md
Specifies a list of values, which will be rendered as a dropdown list in the UI form, for the user to select from. Only values in this list are considered valid for this property.
```

docs/schema.md
jprzychodzen: I'm not sure - this is an example how to use enum, other parts of documentation are not type specific (eg. `minimum` and `maximum`). To keep the same style in `enum`, I prefer not to add type explicitly for only this property. I think it would look strange.

docs/schema.md
jprzychodzen: Done

PR #355, opened 2019-05-21
Install from version file

huyhg:
The version file implies both deployer image and
version repo.

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -19,20 +19,20 @@ set -eo pipefail
 for i in "$@"
 do
 case $i in
-  --deployer=*)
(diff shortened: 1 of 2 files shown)


scripts/install
```diff
@@ -19,20 +19,20 @@ set -eo pipefail
...
 case $i in
-  --deployer=*)
-    deployer="${i#*=}"
-    shift
-    ;;
   --parameters=*)
     parameters="${i#*=}"
     shift
     ;;
+  --deployer=*)
```
vcanaa: Please keep the order if there is changing it was not intentional.

PR #359, opened 2019-05-24
Generate passwords and TLS certificates at provisioning time

huyhg:
These values have to remain the same across updates.

### DIFF
--- a/marketplace/deployer_util/provision.py
+++ b/marketplace/deployer_util/provision.py
@@ -21,6 +21,7 @@ from argparse import ArgumentParser
 import yaml
 
 import config_helper
+import property_generator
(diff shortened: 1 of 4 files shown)


marketplace/deployer_util/provision.py
```diff
@@ -99,6 +100,10 @@ def process(schema, values, deployer_image, deployer_entrypoint, version_repo,
     elif prop.xtype == config_helper.XTYPE_INGRESS_AVAILABLE:
       # TODO: Really populate this value.
```
vcanaa: Since you are touching this file, can you link to #360

TODO(#360): Really populate this value.

PR #363, opened 2019-06-05
First pass at integrating billing metering into a Kubernetes app.

jvolkman:


docs/billing-integration.md
```diff
@@ -0,0 +1,273 @@
...
+data:
+  config.yaml: |
+    # The identity section contains authentication information used
+    # by the agent.
+    identities:
+    - name: gcp
+      gcp:
+        # This parameter accepts a base64-encoded JSON service
+        # account key. The value comes from the reporting secret.
+        encodedServiceAccountKey: $AGENT_ENCODED_KEY
```
huyhg: I don't remember if this will get substituted with empty string by `envsubst`. I _believe_ we fixed it so that unrecognized variables don't get substituted.

Regardless, I think it's a good idea to highlight that this isn't intended to be substituted. Maybe changing `$name` to something helm-like `{{ .name }}`? We should change all env variables in this snippet as well.

docs/billing-integration.md
```diff
@@ -0,0 +1,273 @@
...
+          mountPath: /etc/ubbagent
+      volumes:
+      - name: ubbagent-config
+        configMap:
+          name: ubbagent-config
+```
+
+`schema.yaml` excerpt:
+```yaml
+application_api_version: v1beta1
```
huyhg: rm. This is because this field is not applicable in schema v2.

docs/billing-integration.md
```diff
@@ -0,0 +1,273 @@
...
+properties:
+  name:
+    type: string
+    x-google-marketplace:
+      type: NAME
+  imageMyApp:
+    type: string
+    default: $REGISTRY:$TAG
```
huyhg: Replace variables with real values.

These are intended to be literal image URLs. We just happen to use substitutions in our example app for automation, which requires `envsubst` invocation in the Dockerfile.

docs/billing-integration.md
```diff
@@ -0,0 +1,273 @@
...
+    type: string
+    x-google-marketplace:
+      type: REPORTING_SECRET
+```
+
+The important aspects of the above examples include:
+* The metering agent's config is included as a value named `config.yaml` within a config-map.
+* This config-map is mounted as a volume at `/etc/ubbagent` in the ubbagent container, making the config file available to the agent at `/etc/ubbagent/config.yaml`. This value is passed to the agent sidecar using the special `$AGENT_CONFIG_FILE` environment variable.
+* The agent is instructed to listen on port `4567` using the special `$AGENT_LOCAL_PORT` environment variable. Note that this value must be quoted since environment variable values must be strings.
+* The name of the reported secret created at deployment time is stored in the `$reportingSecret` parameter.
```
huyhg: drop `$`

docs/billing-integration.md
jvolkman: Added an explanation.

PR #368, opened 2019-06-13
Adding image_pull_secret property to verifier

vcanaa:
This property can be used to tell the verifier what imagePullSecret to use.

This change is needed by on-prem verification.

### DIFF
--- a/scripts/verify
+++ b/scripts/verify
@@ -98,7 +98,7 @@ kubectl create namespace "$NAMESPACE"
 
 if [[ -n "$image_pull_secret" ]]; then
   echo "Copying secret/$image_pull_secret from namespace/$default to namespace/$NAMESPACE."
-  kubectl get secret "$image_pull_secret" --namespace=default --export -o=yaml | kubectl apply --namespace="$NAMESPACE" -f -
+  kubectl get secret "$image_pull_secret" --namespace=default --export -o=yaml | kubectl create --namespace="$NAMESPACE" -f -
   kubectl get secret "$image_pull_secret" --namespace="$NAMESPACE"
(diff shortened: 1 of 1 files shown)


scripts/verify
```diff
@@ -92,6 +96,13 @@ echo "INFO Parameters: $parameters"
 echo "INFO Creates namespace \"$NAMESPACE\""
 kubectl create namespace "$NAMESPACE"
 
+if [[ -n "$image_pull_secret" ]]; then
+  echo "Copying secret/$image_pull_secret from namespace/$default to namespace/$NAMESPACE."
+  kubectl get secret "$image_pull_secret" --namespace=default --export -o=yaml | kubectl apply --namespace="$NAMESPACE" -f -
```
huyhg: use `create` instead of `apply` to avoid clobbering existing resources

scripts/verify
```diff
@@ -40,6 +40,10 @@ case $i in
     additional_deployer_role="${i#*=}"
     shift
     ;;
+  --image_pull_secret=*)
+    image_pull_secret="${i#*=}"
```
huyhg: per offline discussion: image_pull_secret_name

scripts/verify
```diff
@@ -146,6 +157,7 @@ echo "INFO Initializes the deployer container which will deploy all the applicat
   --deployer="$deployer" \
   --parameters="$parameters" \
   --entrypoint='/bin/deploy_with_tests.sh' \
+  --image_pull_secret="$image_pull_secret" \
```
huyhg: we shouldn't need this if the namespace is already annotated?

scripts/verify
vcanaa: Per further discussion keeping image_pull_secret to keep consistent with the same parameter in provision.py

PR #369, opened 2019-06-19
Add gsutil to doctor

wgrzelak:


scripts/doctor.py
```diff
@@ -50,7 +50,8 @@ def main():
       gcloud_project=(check_gcloud_default_project, ['gcloud_login']),
       kubectl=check_kubectl,
       kubectl_nodes=(check_kubectl_nodes, ['kubectl']),
-      crd=(check_crd, ['kubectl_nodes']))
+      crd=(check_crd, ['kubectl_nodes']),
+      gsutil=check_gsutil)
```
huyhg: this should be after `['gcloud_login']`

scripts/doctor.py
```diff
@@ -48,10 +48,10 @@ def main():
       gcloud=check_gcloud,
       gcloud_login=(check_gcloud_login, ['gcloud']),
       gcloud_project=(check_gcloud_default_project, ['gcloud_login']),
+      gsutil=check_gsutil,
```
huyhg: when I said after, I meant to include gcloud_loging as a dependency. The order here doesn't matter. This should be written as:

```python
gsutil=(check_gsutil, ['gcloud_login'])
```

wgrzelak: https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/375

PR #370, opened 2019-06-19
Include error summary session at the bottom of verify script logs

vcanaa:
We now append a session at the end of the log with the error summary:

```
=== ERROR SUMMARY ===
Some errors
Some more errors
=====================
```

This is intended to make the errors easier to spot.


vcanaa:
\gcbrun

scripts/verify
```diff
@@ -55,6 +55,75 @@ done
...
 
+function delete_namespace() {
+  if [[ -z "$namespace_deleted" ]]; then
+    echo "INFO Collecting events for namespace \"$NAMESPACE\""
+    kubectl get events --namespace="$NAMESPACE" || echo "ERROR Failed to get events for namespace \"$NAMESPACE\""
+    error_events="$(kubectl get events --namespace="$NAMESPACE" --field-selector="type!=Normal" || echo "ERROR Failed to get events for namespace \"$NAMESPACE\"")"
+
+    if [[ ! -z $deployer_name ]]; then
+      echo "INFO Collecting logs for deployer"
+      deployer_logs="$(kubectl logs "jobs/$deployer_name" --namespace="$NAMESPACE" || echo "ERROR Failed to get logs for deployer $deployer_name")"
```
huyhg: should probably add `--tail 9999` just to be safe

scripts/verify
```diff
@@ -55,6 +55,75 @@ done
 [[ -z "$parameters" ]] && parameters="{}"
 [[ -z "$wait_timeout" ]] && wait_timeout=600
 
+function delete_namespace() {
+  if [[ -z "$namespace_deleted" ]]; then
+    echo "INFO Collecting events for namespace \"$NAMESPACE\""
+    kubectl get events --namespace="$NAMESPACE" || echo "ERROR Failed to get events for namespace \"$NAMESPACE\""
+    error_events="$(kubectl get events --namespace="$NAMESPACE" --field-selector="type!=Normal" || echo "ERROR Failed to get events for namespace \"$NAMESPACE\"")"
```
huyhg: remove the `||`, because `error_events` would contain the error message in case of failure.

scripts/verify
```diff
@@ -55,6 +55,75 @@ done
...
+    error_summary="$error_summary$1\n"
+  fi
+  clean_resources
+  exit 1
+}
+
+function print_summary() {
+  if [[ -n "$error_events" ]]; then
+    error_summary="${error_summary}Error events found in namespace \"$NAMESPACE\"\n"
+    error_summary="$error_summary$error_events\n"
```
huyhg: beconsistent, use `${}` every where

scripts/verify
```diff
@@ -55,6 +55,75 @@ done
...
+    echo -e "$1"
+    error_summary="$error_summary$1\n"
+  fi
+  clean_resources
+  exit 1
+}
+
+function print_summary() {
+  if [[ -n "$error_events" ]]; then
+    error_summary="${error_summary}Error events found in namespace \"$NAMESPACE\"\n"
```
huyhg: It's safer to use `printf` directly here to turn `\n` into new lines, rather than `echo -e` on the entire log

scripts/verify
```diff
@@ -55,6 +55,75 @@ done
...
+  fi
+
+  if [[ -n "$deployer_errors" ]]; then
+    error_summary="${error_summary}Error traces found in deployer\n"
+    error_summary="$error_summary$deployer_errors\n"
+  fi
+
+  if [[ -n "$error_summary" ]];then
+    echo "=== ERROR SUMMARY ==="
+    echo -e "$error_summary"
```
huyhg: remove `-e`--see my comment above

scripts/verify
vcanaa: I want it to show up in the logs in case we fail to get events.

PR #373, opened 2019-06-25
 Fix error summary printing.

vcanaa:


scripts/verify
```diff
@@ -98,17 +98,17 @@ function clean_and_exit() {
...
-    error_summary="$(printf ${error_summary}Error traces found in deployer\\n${deployer_errors}\\n)"
+    error_summary="${error_summary}\nError traces found in deployer\n${deployer_errors}\n"
   fi
 
   if [[ -n "$error_summary" ]];then
-    echo "=== ERROR SUMMARY ==="
-    echo "$error_summary"
-    echo "====================="
+    printf "=== ERROR SUMMARY ===\n"
+    printf "$error_summary"
```
huyhg: maybe remove the trailing `\n` in `error_summary` and just revert to the previous `echo` version.

Alternatively, use `echo -n` to avoid adding a newline. But I think the above suggestion is cleaner.

PR #374, opened 2019-06-26
Print error summary from file

vcanaa:


scripts/verify
```diff
@@ -90,25 +92,29 @@ function clean_resources() {
...
-  if [[ -n "$error_summary" ]];then
-    printf "=== ERROR SUMMARY ===\n"
-    printf "$error_summary"
-    printf "=====================\n"
+  if [[ -f "$error_summary_path" ]]; then
+    echo "=== ERROR SUMMARY ==="
+    cat "$error_summary_path"
+    echo "====================="
   fi
 }
```
huyhg: on L173 it was incorrectly written as `namespace/$default`. Can you remove the `$`?

scripts/verify
```diff
@@ -55,6 +55,8 @@ done
 [[ -z "$parameters" ]] && parameters="{}"
 [[ -z "$wait_timeout" ]] && wait_timeout=600
 
+error_summary_path="error_summary.txt"
```
huyhg: can we fix the location instead of having this in a relative path, like `/tmp/error_summary`? or create a temporary file and store the value here using `mktemp`

PR #381, opened 2019-07-16
Provision deployer permissions defined in schema.yaml

eshiroma:
/gcbrun


eshiroma:
Note: this change should not be released until the TODOs to cleanup the cluster-scoped IAM resources are resolved (or else deployments with such resources will leave behind the ClusterRoles and ClusterRoleBindings provisioned for the deployer, which will be applied to a service account with the same name).

PR #383, opened 2019-07-16
Validate default value for IMAGE properties in schema v1

eshiroma:
Although unused in the final deployment, these values are key for the publishing process since they indicate which staging GCR repo values to copy.

/gcbrun

### DIFF
--- a/marketplace/deployer_util/config_helper.py
+++ b/marketplace/deployer_util/config_helper.py
@@ -101,8 +101,8 @@ class Schema:
 
   def __init__(self, dictionary):
     self._x_google_marketplace = _maybe_get_and_apply(
-        dictionary,
-        'x-google-marketplace', lambda v: SchemaXGoogleMarketplace(v))
+        dictionary, 'x-google-marketplace',
+        lambda v: SchemaXGoogleMarketplace(v))
 
     self._required = dictionary.get('required', [])
     self._properties = {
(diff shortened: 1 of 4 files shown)


marketplace/deployer_util/config_helper.py
```diff
@@ -669,10 +669,19 @@ def application_create(self):
...
   """Accesses IMAGE and DEPLOYER_IMAGE properties."""
 
-  def __init__(self, dictionary):
+  def __init__(self, dictionary, default):
     self._split_by_colon = None
     self._split_to_registry_repo_tag = None
 
+    if not default:
+      raise InvalidSchema('default image value must be specified')
+    if not default.startswith('gcr.io'):
```
vcanaa: Should we be concerned about marketplace.gcr.io or region based endpoints (ex: us-west1.gcr.io)?

vcanaa: I advise against this particular check for gcr.io, since I don't think we need it. It does conform with the data we have seen so far and it is easy to change, so I am fine either way.

eshiroma: Not yet. Our system currently **requires** partner staging GCR repositories to be in `gcr.io` ([container image solution documentation](https://cloud.google.com/marketplace/docs/partners/container-setup#tag_and_push_images); Kubernetes App documentation should have the same notice added since it has the same requirement)

We can update this check if we move to supporting other regions for staging repos.

PR #384, opened 2019-07-17
Add section on IMAGE schema.yaml property

eshiroma:


docs/schema.md
```diff
@@ -287,6 +287,15 @@ It defines how this object will be handled. Each type has a different set of pro
 
 ---
 
+### type: IMAGE
```
vcanaa: This only applies for schema v1. Lets add this session here and have a working item to include schema v2 instructions.

vcanaa: Created #387.

eshiroma: +1 Thanks for tracking!

PR #385, opened 2019-07-17
Accept a canonical_image_pull_secret to create and apply to all KSAs during install

eshiroma:
Intended to replace the existing image_pull_secret (referencing an annotated image pull secret in the deployment namespace).

/gcbrun


eshiroma:
Closing since the image pull secret use case is no longer in focus.

PR #390, opened 2019-08-09
Enforce partner to run 'mpdev verify' command without parameters.

vcanaa:
As explain in the mpdev doc, parameters for verification should be passed in using the default values in the schema file.


scripts/verify
```diff
@@ -183,17 +183,34 @@ if [[ "$istio" = "enabled" ]]; then
...
 
+reporting_secret_key="$(extract_schema_key.py \
+  --schema_file=/data/schema.yaml \
+  --type=REPORTING_SECRET \
+  | head -n 1)"
+
+# Stitch in fake REPORTING_SECRET parameter if applicable.
+if [[ -n "$reporting_secret_key" ]]; then
+  if [[ "$(echo "$parameters" | jq ".$reporting_secret_key")" == "null" ]]; then
+    kubectl apply -f "/data/fake-reporting-secret-manifest.yaml" --namespace="$NAMESPACE"
```
huyhg: is there a way to dynamically generate the secret name instead?

scripts/verify
```diff
@@ -183,17 +183,34 @@ if [[ "$istio" = "enabled" ]]; then
...
+    parameters=$(echo "$parameters" \
+      | jq \
+        --arg reporting_secret_key "$reporting_secret_key" \
```
huyhg: I don't understand the purpose of `--arg` here. Is it to work around the single quoted text not expanding env var value?

I think the following should work:

```
| jq "$(printf '.%s = "fake-reporting-secret-manifest"' "$reporting_secret_key")"
```

Tried in a shell the following:

```bash
key="mykey"
$ echo '{"a": "b"}' | jq "$(printf '.%s = "value"' "$key")" 
{
  "a": "b",
  "mykey": "value"
}
```

scripts/verify
huyhg: add a log entry

tests/marketplace/deployer_envsubst_base/full/app.Makefile
```diff
@@ -77,14 +70,12 @@ app/install:: app/build \
...
                    .build/var/APP_DEPLOYER_IMAGE \
                    .build/var/APP_PARAMETERS \
-                   .build/var/APP_TEST_PARAMETERS \
                    .build/var/MARKETPLACE_TOOLS_TAG \
 	           | .build/app/dev
 	$(call print_target)
 	.build/app/dev \
 	    /scripts/install \
 	        --deployer='$(APP_DEPLOYER_IMAGE)' \
-	        --parameters='$(call combined_parameters)' \
```
huyhg: this should be retained and changed to `APP_PARAMETERS`

tests/marketplace/deployer_envsubst_base/full/app.Makefile
```diff
@@ -103,29 +80,25 @@ app/uninstall: .build/var/APP_DEPLOYER_IMAGE \
...
             .build/var/APP_PARAMETERS \
-            .build/var/APP_TEST_PARAMETERS \
             .build/var/MARKETPLACE_TOOLS_TAG \
             | .build/app/dev
 	$(call print_target)
 	.build/app/dev \
 	    /scripts/verify \
-	          --deployer='$(APP_DEPLOYER_IMAGE)' \
-	          --parameters='$(call combined_parameters)'
+	          --deployer='$(APP_DEPLOYER_IMAGE)'
```
huyhg: can you re-indent this using 4 spaces only (it's 6 right now)?

PR #396, opened 2019-08-26
Clean up provisioned custom deployer permissions after deployment

eshiroma:
Grants the deployer permission to list IAM resources and to delete the ones provisioned for deployment (by name). Deletes the provisioned resources after deployment, and then the ClusterRoleBinding that grants permission to perform these deletions.

Note: Should not be released until the UI has been updated (and released) to grant the new cleanup permissions and set the corresponding labels.

/gcbrun


eshiroma:
The UI change is in its final rollout stages.

Note: the integration test fails on CloudBuild since the test only grants cluster-admin within the namespace, and cluster-scoped permissions are required. Test passes locally when user (tester) has cluster-admin.
This issue is solved by https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/399

huyhg:
Removing the reviewers for now. Please add back when this PR is ready for review again.

eshiroma:
Hesitant to preserve given the many changes since. Note that the label was changed to `deployer-rbac.marketplace.cloud.google.com`

PR #399, opened 2019-08-28
Create a new cluster for CloudBuild tests instead of using a predefined one

eshiroma:
Improve the hermetic aspect of tests, and enable testing apps with cluster-level resources.


eshiroma:
DO NOT MERGE until a backup cluster cleanup mechanism is implemented.

PR #417, opened 2019-10-18
Only apply the App OwnerRef to namespaced resources

eshiroma:
Prevent applying OwnerRefs that do not comply with the [Kubernetes spec](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents) (which forbids namespaced resources owning cluster-scoped resources).

/gcbrun


eshiroma:
(Also adds unit tests for set_resource_ownership)

PR #424, opened 2019-11-05
Validate schema publishedVersion value

eshiroma:
/gcbrun

### DIFF
--- a/marketplace/deployer_util/config_helper.py
+++ b/marketplace/deployer_util/config_helper.py
@@ -23,6 +23,7 @@ import sys
 import yaml
 
 NAME_RE = re.compile(r'[a-zA-z0-9_\.\-]+$')
+SEMVER_RE = re.compile(r'^[0-9]+\.[0-9]+\..+$')
 
 XGOOGLE = 'x-google-marketplace'
 XTYPE_NAME = 'NAME'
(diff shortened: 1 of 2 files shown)


marketplace/deployer_util/config_helper.py
```diff
@@ -23,6 +23,7 @@
 import yaml
 
 NAME_RE = re.compile(r'[a-zA-z0-9_\.\-]+$')
+SEMVER_RE = re.compile(r'^[0-9]+\.[0-9]+\..+$')
```
eshiroma: Python doesn't appear to have any built-in semver modules, and this covers the main cases (e.g. partners only provide a minor version).

marketplace/deployer_util/config_helper.py
```diff
@@ -23,6 +23,7 @@
 import yaml
 
 NAME_RE = re.compile(r'[a-zA-z0-9_\.\-]+$')
+SEMVER_RE = re.compile(r'^[0-9]+\.[0-9]+\.([^.]+)$')
```
huyhg: This doesn't allow metadata (the part after the dash, which C2D is using).

Can we use the one from https://semver.org/

```
^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
```

eshiroma: The existing regex does allow for metadata:
```
$ python
Python 2.7.17rc1 (default, Oct 10 2019, 10:26:01)
[GCC 8.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> SEMVER_RE = re.compile(r'^[0-9]+\.[0-9]+\.([^.]+)$')
>>> SEMVER_RE.match('1.0.0-metadata')
<_sre.SRE_Match object at 0x7f45ca951f30>
>>>
```

but will be too lenient on other semver requirements.

Follow-up to use the suggested regex is in https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/426

PR #433, opened 2019-12-04
"mpdev script/install" should not set cluster level resource owerRef which has GC issue

rmgogogo:
"mpdev script/install" should not set cluster level resource owerRef which has GC issue


eshiroma:
Replaced with [PR #435](https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/435). Planning to create a new release by end of day.

PR #436, opened 2019-12-04
Validate SERVICE_ACCOUNT roles

eshiroma:
Ensure that SERVICE_ACCOUNT properties are validly formed to catch errors early. Otherwise, issues only surface when the property is parsed in the UI.

/gcbrun

### DIFF
--- a/marketplace/deployer_util/config_helper.py
+++ b/marketplace/deployer_util/config_helper.py
@@ -798,6 +798,12 @@ class SchemaXServiceAccount:
                                                      rule.get('apiGroups')):
             raise InvalidSchema("Missing or empty apiGroups in rules. "
                                 "Did you mean [\"v1\"] or [\"*\"]?")
+          if not rule.get('resources') or not filter(lambda x: x,
+                                                     rule.get('resources')):
+            raise InvalidSchema('Missing or empty resources in rules.')
+          if not rule.get('verbs') or not filter(lambda x: x,
+                                                 rule.get('verbs')):
+            raise InvalidSchema('Missing or empty verbs in rules.')
       else:
         raise InvalidSchema('rulesType must be one of PREDEFINED or CUSTOM')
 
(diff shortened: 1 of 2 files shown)


marketplace/deployer_util/config_helper.py
```diff
@@ -778,6 +778,28 @@ class SchemaXServiceAccount:
...
+        if not role.get('rulesFromRoleName'):
+          raise InvalidSchema('Missing rulesFromRoleName for PREDEFINED role')
+      elif role.get('rulesType') == 'CUSTOM':
+        if role.get('rulesFromRoleName'):
+          raise InvalidSchema(
+              'rulesFromRoleName can only be used with rulesType PREDEFINED')
+        if not role.get('rules'):
+          raise InvalidSchema('Missing rules for CUSTOM role')
+        for rule in role.get('rules', []):
+          if rule.get('nonResourceURLs'):
```
huyhg: probably require `resources` as well, since otherwise the rules don't really have effects

eshiroma: Good call, added check for `resources` and `verbs`. I had forgotten to do so since no one's made that mistake yet - now no one will :)

marketplace/deployer_util/config_helper.py
```diff
@@ -778,6 +778,34 @@ class SchemaXServiceAccount:
...
+            raise InvalidSchema(
+                'Only attributes for resourceRules are supported in rules')
+          if not rule.get('apiGroups') or not filter(lambda x: x,
+                                                     rule.get('apiGroups')):
+            raise InvalidSchema("Missing or empty apiGroups in rules. "
+                                "Did you mean [\"v1\"] or [\"*\"]?")
```
schottsfired: Hi @eshiroma, this rule broke our CI command (`mpdev verify` ..) because we have empty `apiGroups`:
https://github.com/cloudbees/core-google-launcher/blob/master/schema-test.yaml#L39-L156

Per K8s documentation, it looks like `""` is an acceptable value for `apiGroups`:
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole

So, is this validation really needed? Is there something else we might be missing?

Thanks!

cc @Vlatombe @jglick

PR #437, opened 2019-12-05
Update docs for the now-optional primary app image

eshiroma:
Should not be merged until server validation relaxations are fully released.

/gcbrun


eshiroma:
Replacing with #471 for a cleaner commit history

PR #438, opened 2019-12-06
Use latest Helm version - 2.16.1, which supports ternary function.

danielmkn:
Helm 2.8.1 doesn't support ternary in YAML files. 
In this example, used by Postgres chart (https://github.com/helm/charts/blob/master/stable/postgresql/templates/statefulset.yaml, rows 121 and 202) ternary won't be parsed and deployment will fail. 
```value: {{ ternary "true" "false" .Values.image.debug | quote }}```

Error message: "Error: parse error in "artifactory-ha/charts/postgresql/templates/statefulset.yaml": template: artifactory-ha/charts/postgresql/templates/statefulset.yaml:121: function "ternary" not defined"

To resolve this problem we need to use the latest version of Helm in the Deployer dockerfile.


danielmkn:
@googlebot I signed it!

danielmkn:
@huyhg, please review

Issue #443, opened 2019-12-12
ServiceAccounts do not own ClusterRoles in 0.9.6


DTMad:
Hey!

When deploying a deployer with a schema.yaml file which consists ServiceAccounts with ClusterRoles underneath with mpdev the ServiceAccounts won't own the ClusterRoles after creation.
The problem is not the 'does not own' but the ClusterRoles do not get created at all.

This most probably happens because of that:
https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/9a00f6384e6fa135cdd8ea7d2ba695c8b8246166/marketplace/deployer_util/set_ownership.py#L143
The 'continue' there skips the 'to_be_dumped.append(resource)' underneath.

Occurs in v0.9.6
Does not occur in v0.9.5

**mpdev output in v0.9.6:**
`Using /opt/kubectl/1.13/kubectl (server=1.13)
Using /opt/kubectl/1.13/kubectl (server=1.13)
application.app.k8s.io/dynatrace-oneagent-operator created
Using /opt/kubectl/1.13/kubectl (server=1.13)
INFO Application 'xxx' owns 'ServiceAccount/xxx-operatorserviceaccount'
INFO Application 'xxx' owns 'Role/xxx:operatorServiceAccount-r0'
INFO Application 'xxx' owns 'RoleBinding/xxx:operatorServiceAccount-rb0'
INFO Application 'xxx' does not own cluster-scoped 'ClusterRole/xxx:xxx:operatorServiceAccount-r0'
INFO Application 'xxx' does not own cluster-scoped 'ClusterRoleBinding/xxx:xxx:operatorServiceAccount-rb0'  INFO Application 'xxx' owns 'ServiceAccount/xxx-crdserviceaccount'
INFO Application 'xxx' does not own cluster-scoped 'ClusterRole/xxx:crdServiceAccount-r0'
INFO Application 'xxx' does not own cluster-scoped 'ClusterRoleBinding/xxx:xxx:crdServiceAccount-rb0'       INFO Application 'xxx' owns 'ServiceAccountxxx'
INFO Application 'xxx' owns 'Role/xxx:oneagentServiceAccount-r0'
INFO Application 'xxx' owns 'RoleBinding/xxx:oneagentServiceAccount-rb0'
INFO Application 'xxx' owns 'ServiceAccount/xxx-deployer-sa'
[...]

PR #447, opened 2019-12-16
Create cluster-scoped resources (without OwnerRefs)

eshiroma:
To address [Issue #443](https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/issues/443).

/gcbrun


marketplace/deployer_util/set_ownership.py
```diff
@@ -140,6 +140,7 @@ def dump(outfile, resources, included_kinds, app_name, app_uid,
       # https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents
       log.info("Application '{:s}' does not own cluster-scoped '{:s}/{:s}'",
                app_name, resource["kind"], resource["metadata"]["name"])
+      to_be_dumped.append(resource)
```
huyhg: optional: to make this clearer (and avoid something like this), it's probably better to extract the content of the `for` loop into a transforming function, which could live within the scope of this `dump()` function, and rewrite this as:

```python
to_be_dumped = [maybe_assign_ownership(resource) for resource in resources]
```

eshiroma: Done.

PR #449, opened 2019-12-16
Validate partner and solution ID

huyhg:
Also make an update the dev README file. Due to changes in gcloud and kubectl a few months ago, it's now required to use application default credentials. Using a Service Account instead of user credential for gcloud authentication is generally recommended; it also fixes an annoying issue with 24 hour re-authentication requirement for Googlers.


huyhg:
I just merged in `master`. There were a few conflicting changes. As a result, it's not a clean diff when we compare this change against its previous incarnation in the `master` branch: https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/compare/292696d..huyhg/fix

Changes shown up here include changes that have been committed to master. A 3-way diff would have made this better but I don't have a way of doing that in GitHub.

README.md
```diff
@@ -25,6 +25,66 @@ We follow [Google's coding style guides](https://google.github.io/styleguide/).
...
 
+## Setting up
+
+### Log in `gcloud` with a Service Account
+
+Instead of using your personal credential to log in, it's recommended
+to use a Service Account instead.
+
+A new Service Account and proper permissions can be created using the
+following commands. `PROJECT-ID` is the (non-numeric) name of your
```
eshiroma: nit: maybe change "name" to "identifier", since "project name" is actually another separate field.

tests/py/runtests.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -eox pipefail
```
eshiroma: Looks like the `x` might make tests fail? Is there another way for us to print the shell commands for future debugging?

huyhg: It doesn't cause the tests to fail--it's just unnecessarily verbose. What the user cares about is the output reports, not the command to generate the reports itself.

We don't use `-x` in other scripts, either. To debug the Makefiles and build files themselves, I think it's reasonable to make temporary changes to them.

huyhg:
Ready for a final look.

PR #450, opened 2019-12-17
Log all applied manifests in install

eshiroma:
Increase logging for failure debugging.

/gcbrun


eshiroma:
Includes saving logs to the user's /tmp directory for improved debugging when running `install` manually. These logs will be automatically cleaned up for Verification (CloudBuild) runs.

/gcbrun

eshiroma:
Instead removing /tmp manifests (best-effort) to prevent orphaned resource buildup, as the user will already have all contents printed in the console.

/gcbrun

PR #457, opened 2020-01-06
Support service account kind as owner in resources util

eshiroma:
/gcbrun


eshiroma:
Replacing with #459 which has a cleaner branch history

Issue #347, opened 2019-05-08
Deployer job failed to delete deployer-sa 


depohmel:
Error message:
```
+ kubectl delete --namespace=g-launcher-zbnuszsdu8zentbx --filename=-
serviceaccount "k10-deployer-sa" deleted
error: You must be logged in to the server (error when deleting "STDIN": Unauthorized)
+ handle_failure
```

deployer image: 
```
FROM gcr.io/cloud-marketplace-tools/k8s/deployer_helm
Digest: sha256:e94ad791c3b0441f84a801438030489d3a7c89d9ff7b88ee97321cfceb303944
Status: Downloaded newer image for gcr.io/cloud-marketplace-tools/k8s/deployer_helm:latest
```

The application was deployed successfully but marked as failed after failed cleanup attempt.

It's probably a good idea that cleanup code will not fail application status. 
 
cluster details:
```
Master version 1.12.7-gke.10
Client certificate
Disabled
Binary Authorization
Disabled
Kubernetes alpha features
Disabled
Total size
3
Region
us-west1
Node zones
us-west1-b
us-west1-a
us-west1-c
Network
default
Subnet
default
VPC-native (alias IP)
Disabled
Intranode visibility
Disabled
Stackdriver Logging
Disabled
Stackdriver Monitoring
Disabled
Private cluster
Disabled
Master authorized networks
Disabled
Network policy
Enabled
Legacy authorization
Disabled
Maintenance window
Any time
Cloud TPU
Disabled
Application-layer Secrets Encryption
Disabled
Node auto-provisioning
Disabled
Vertical Pod Autoscaling
Disabled
GKE usage metering
Disabled  
```

eshiroma:
**Proposal: only explicitly delete the ServiceAccount, and let GC handle the Role and RoleBinding.**

Thanks for the report, Ilya.

This is likely caused by the fact that the deployer ServiceAccount (KSA) is deleted in the same kubectl request as the Role and RoleBinding that grant it permissions to act. Deletion order is nondeterministic.
From [clean_iam_resources.sh](https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/master/marketplace/deployer_util/clean_iam_resources.sh#L29):
```
kubectl delete --namespace="$NAMESPACE" \
  ServiceAccount,Role,RoleBinding ...
```

Instead, the deployer KSA should own the Role and RoleBinding, and only the KSA should be explicitly deleted:
[...]

PR #458, opened 2020-01-06
Make OwnerRefs handle cleanup of deployer's RBAC resources

eshiroma:
Set the deployer service account as owner of it's component resources (i.e. Role and RoleBinding) during install (already done during UI deployment), then only delete the deployer ServiceAccount after deployment, rather than also explicitly deleting the Role and Rolebinding.

Addresses #347:
Partners must update their base deployer to take advantage of this change, which is to only delete the deployer service account after deployment instead of also attempting to delete its dependent resources to avoid deletion timing issues.

/gcbrun


eshiroma:
Replaced by #460

PR #460, opened 2020-01-07
Make OwnerRefs handle cleanup of deployer's RBAC resources

eshiroma:
Set the deployer service account as owner of it's component resources (i.e. Role and RoleBinding) during install (already done during UI deployment), then only delete the deployer ServiceAccount after deployment, rather than also explicitly deleting the Role and Rolebinding.

Addresses #347:
Partners must update their base deployer to take advantage of this change, which is to only delete the deployer service account after deployment instead of also attempting to delete its dependent resources to avoid deletion timing problems.

/gcbrun

### DIFF
--- a/marketplace/deployer_util/clean_iam_resources.sh
+++ b/marketplace/deployer_util/clean_iam_resources.sh
@@ -20,6 +20,8 @@ set -eox pipefail
 [[ -z "$NAMESPACE" ]] && echo "NAMESPACE must be set" && exit 1
 
 # Delete the service account (which owns its RBAC objects) by label.
+# Update _DEPLOYER_OWNED_KINDS in set_ownership.py to delete additional
+# resource types besides Role and RoleBinding.
 # Note that only resources of a one-shot deployer have this label.
 # Resources of a KALM-managed deployer have
 # app.kubernetes.io/component=kalm.marketplace.cloud.google.com label.
--- a/marketplace/deployer_util/clean_iam_resources.sh
+++ b/marketplace/deployer_util/clean_iam_resources.sh
@@ -19,12 +19,11 @@ set -eox pipefail
 [[ -z "$NAME" ]] && echo "NAME must be set" && exit 1
 [[ -z "$NAMESPACE" ]] && echo "NAMESPACE must be set" && exit 1
 
-# Delete the service account and RBAC objects by labels.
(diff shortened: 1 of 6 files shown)


marketplace/deployer_util/clean_iam_resources.sh
```diff
@@ -19,12 +19,11 @@ set -eox pipefail
 [[ -z "$NAME" ]] && echo "NAME must be set" && exit 1
 [[ -z "$NAMESPACE" ]] && echo "NAMESPACE must be set" && exit 1
 
-# Delete the service account and RBAC objects by labels.
+# Delete the service account (which owns its RBAC objects) by label.
```
vcanaa: Does it mean that only RBAC objects owned by the service account will be deleted?

scripts/install
```diff
@@ -114,14 +114,30 @@ app_uid=$(kubectl get "applications.app.k8s.io/$name" \
   --namespace="$namespace" \
   --output=jsonpath='{.metadata.uid}')
 
+# Create ServiceAccount instance.
+deployer_name="$(make_dns1123_name.py --name="${name}")-deployer-sa"
```
vcanaa: Can you add a comment that this needs to match the deployer name that is provisioned in provision.py

vcanaa: Should the variable name be service_account_name?

eshiroma: Instead of trying to coordinate naming schemes, I've changed provision.py to require the caller to pass in the desired deployer service account name (it's fine if this KSA doesn't exist yet).

---

Often times we say 'deployer' to refer to the deployment service account. Changed variable name to `deployer_service_account_name` to be more clear.

marketplace/deployer_util/clean_iam_resources.sh
eshiroma: Yes.
That is the intention of this `kubectl delete` call - to delete the deployer's RBAC objects (hence the label), which will now be owned by the deployer service account.

Added a comment to point to the kinds list in case other types need to be deleted later.

eshiroma: (Note that this script is for deleting resources associated with the deployer right after deployment completes, not for deleting application resources when the app is deleted. Other application resources are still deleted with the app, as they are owned by the Application)

PR #467, opened 2020-01-14
Setting minimum required cpu and memory values for the deployer

AndreyChoi:
The deployer will encounter errors if there aren't enough resources
for it to run successfully. This sets the minimum values for the
deployer:

- cpu: 1m
- memory: 2Mi


marketplace/deployer_util/provision.py
```diff
@@ -283,18 +283,21 @@ def provision_deployer(schema, app_name, namespace, deployer_image,
...
+            'name': 'deployer',
+            'image': deployer_image,
+            'imagePullPolicy': 'Always',
             'volumeMounts': [{
                 'name': 'config-volume',
                 'mountPath': '/data/values.yaml',
                 'subPath': 'values.yaml',
                 'readOnly': True,
             },],
+            'resources': {
```
eshiroma: Since this is value is used more than once, can it be made into a constant?

PR #469, opened 2020-01-15
Test coverage - bash_util.py and config_helper.py

vcanaa:


marketplace/deployer_util/bash_util.py
```diff
@@ -33,37 +40,18 @@ def __init__(self, cmd, print_call=False, print_result=False):
...
     self._output = None
-    self._error = None
     self._print_call = print_call
     self._print_result = print_result
     self._run()
 
-  def jq(self, query):
-    return self.pipe("jq {}".format(query))
-
-  def pipe(self, cmd):
```
huyhg: we didn't use this?

vcanaa: We didn't in the last version. This should be broken ever since the change to run the comand in the constructor

PR #471, opened 2020-01-16
Update docs for the now-optional primary app image

eshiroma:
/gcbrun

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -259,7 +259,7 @@ In `v1`, images are declared inside the `properties` section, as follows:
 
 ```yaml
 properties:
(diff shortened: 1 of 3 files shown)


docs/schema.md
```diff
@@ -259,7 +259,7 @@ In `v1`, images are declared inside the `properties` section, as follows:
 
 ```yaml
 properties:
-  mainImageName:  # A primary image is required.
```
huyhg: Can you also update L170?

eshiroma: Saw this right after clicking "Merge" - sending a follow-up PR.

PR #481, opened 2020-03-12
Add support and validation for GPU in clusterConstraints

eshiroma:
/gcbrun

PR #482, opened 2020-03-12
Update schema docs for GPU requests

eshiroma:
Only merge after #481 is released.

/gcbrun

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -816,15 +816,18 @@ x-google-marketplace:
 
 Each entry under `resources` is roughly equivalent to a workload in the app.
 
-* `affinity` defines the relationship between the nodes and the replicas.
+*   `requests` defines the desired resource allocations, and is required for each
+    entry.
 
-    * `simpleNodeAffinity` is an affinity definition. It has two types:
+*   `affinity` defines the relationship between the nodes and the replicas.
 
-        * `REQUIRE_ONE_NODE_PER_REPLICA`: The number of nodes must be at least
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -844,6 +847,55 @@ for sufficient available resources, and new clusters will be created with
...
+Currently only Nvidia platforms are supported (requests entry key
+`nvidia.com/gpus`). See the
+[Compute Engine GPU guide](https://cloud.google.com/compute/docs/gpus)
+for a full list of available GPU platforms.
+
+Note that for `resource` entries with GPU requests:
+
+*   `resources` may contain at most one entry with GPU requests
+*   `affinity` and `replicas` are ignored
+    * `cpu` and `memory` may be defined in the same entry, but must be
```
huyhg: Rm. `cpu` and `memory` actually need replicas, otherwise we don't know how to fit.

eshiroma: Done.

docs/schema.md
```diff
@@ -816,15 +816,21 @@ x-google-marketplace:
 
 Each entry under `resources` is roughly equivalent to a workload in the app.
 
-* `affinity` defines the relationship between the nodes and the replicas.
+*   `requests` defines the desired resource allocations, and is required for each
```
huyhg: Style is only a single space after `*`, right?

eshiroma: Updated - I'm not sure what the universal/general best practice is, but leaving as-is (single-space). Internally, autoformatters will correct to having text align at the fourth space.

PR #491, opened 2020-04-14
Fix python lint

huyhg:


marketplace/deployer_util/provision_test.py
```diff
@@ -422,7 +422,7 @@ def generate_schema_v2(self, is_kalm):
           publishedVersionMetadata:
             releaseNote: Initial release
             recommended: True
-          images: {}
+          images: {{}}
```
vcanaa, line 425: Was this intentional? It seems it should match the line below:

kalmSupported: {}

huyhg: `{}` is the format string for `.format`. `{{` is for the literal `{`! Here we do want an empty dict for property `images`.

PR #496, opened 2020-05-08
Mpdev verify wait_timeout should be propagated to install script

vcanaa:


marketplace/deployer_helm_tiller_base/bin/deploy_with_tests.sh
```diff
@@ -34,7 +34,7 @@ patch_assembly_phase.sh --status="Success"
 wait_for_ready.py \
   --name $NAME \
   --namespace $NAMESPACE \
-  --timeout 300
+  --timeout ${WAIT_FOR_READY_TIMEOUT:-1000}
```
huyhg, line 37: There is no way to pass environment variables to the deployer. The only input the deployer operates upon is the schema values. Perhaps we need to introduce special "time out" properties, maybe only for tests?

huyhg: Ah, my comment wasn't correct. The trick was to define `ENV` in the deployer `Dockerfile` directly.

While that works, I think it is inconsistent and hard to discover. I still think defining a new type of property in the schema is the right approach.

vcanaa: Including in the schema.yaml doesn't feel 100% consistent either. The schema is supposed to contain parameters for the installation. But I agree it is easier to discover.

If mpdev has a config file, it would probably reside there.

PR #500, opened 2020-05-14
Restore `cluster-admin` Role to deployer

eshiroma:
To resolve resulting `mpdev verify` failures from #499 

/gcbrun


eshiroma:
Replaced by https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/501

PR #515, opened 2020-07-01
Add permissionExplanation to SERVICE_ACCOUNT property

eshiroma:
To explain why the service account needs the declared permissions, particularly cluster-scoped permissions.

An explicit field is proposed over re-use of the generic `description` property field because:
1. The intended use is different than other fields' `description` usage (as a louder message rather than a more discrete tooltip)
2. The intended content is different than just a description, so using a specific field should help convey the difference

/gcbrun


docs/schema.md
```diff
@@ -559,10 +560,13 @@ properties:
     x-google-marketplace:
       type: SERVICE_ACCOUNT
       serviceAccount:
+        permissionExplanation: >
```
huyhg, line 563: Perhaps simply `description`?

I also think we should make this field required so that the UI always has something to show.

docs/schema.md
```diff
@@ -577,6 +581,10 @@ properties:
             verbs: ['*']
```
huyhg, line 581: nit (above): `apiGroups` entries shouldn't include the version. In the above examples, they should be `apps.kubernetes.io` and `etcd.database.coreos.com`.

eshiroma: Updated.

docs/schema.md
eshiroma: Updated to `description` (hopefully the example is clear enough to distinguish from the property-level `description` field).

Checking within `Schema.validate()` allows the schema to be parsed without it, but not deployed (deployer calls `Schema.validate()`).

eshiroma:
Updated this PR to include:

* Required (by deployer, not dev image) `permissionExplanation` field to `SERVICE_ACCOUNT` property
* (Updated docs and test `schema.yaml` files' `SERVICE_ACCOUNT` properties to include)
* Printing `Schema.validate()` failures in the dev image as a warning. Sample output:
```
...
Using /opt/kubectl/1.16/kubectl (server=1.16)
namespace "apptest-j2b9ffag" deleted
=== SCHEMA VALIDATION WARNING ===
WARNING: Schema is incompatible with the latest deployer, would fail with:

Traceback (most recent call last):
  File "/bin/validate_schema.py", line 35, in <module>
    main()
  File "/bin/validate_schema.py", line 31, in main
    schema_values_common.load_schema_and_validate(args)
  File "/bin/schema_values_common.py", line 53, in memoized_func
    cache[key] = func(*args, **kwargs)
  File "/bin/schema_values_common.py", line 66, in load_schema_and_validate
    return config_helper.Schema.load_yaml_file(parsed_args.schema_file).validate()
  File "/bin/config_helper.py", line 168, in validate
    p.service_account.validate()
  File "/bin/config_helper.py", line 933, in validate
    'SERVICE_ACCOUNT property must have a `description` '
config_helper.InvalidSchema: SERVICE_ACCOUNT property must have a `description` explaining purpose and permission requirements. See docs: https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/master/docs/schema.md#type-service_account
==================================
```

marketplace/deployer_util/schema_values_common.py
```diff
@@ -61,6 +61,11 @@ def load_schema(parsed_args):
   return config_helper.Schema.load_yaml_file(parsed_args.schema_file)
 
 
+@memoize
+def load_schema_and_validate(parsed_args):
+  return config_helper.Schema.load_yaml_file(parsed_args.schema_file).validate()
```
huyhg, line 66: nit: re-use `load_schema`, avoiding memoizing the content twice.

marketplace/deployer_util/config_helper.py
```diff
@@ -160,6 +160,13 @@ def validate(self):
               'schema v2. Images must be declared in the top level '
               'x-google-marketplace.images')
 
+    # Backward-incompatible validations that should only be enforced upon
```
huyhg, line 163: This is the intention of all `validate()` methods. This particular location for the comment is a bit misleading, since "here" does sound like "this line".

Maybe we should add this snippet to `Schema.validate()`, and make all other `validate()` methods reference it?

scripts/verify
```diff
@@ -111,6 +111,14 @@ function print_summary() {
     echo "$deployer_errors" >> "$error_summary_path"
   fi
 
+  if [[ -n "$validation_errors" ]]; then
+    echo "=== SCHEMA VALIDATION WARNING ==="
+    echo "WARNING: Schema is incompatible with the latest deployer, would fail with:"
```
huyhg, line 116: It occurred to me that we should _additionally_ update `deploy_with_tests.sh` to also call schema validation. This should be safe because it's only invoked during `verify`, and if the partner doesn't update their deployer, the new validations wouldn't apply (i.e. verification still passes).

marketplace/deployer_util/schema_values_common.py
eshiroma: Ah, thanks for the suggestion! Fixed.

marketplace/deployer_util/config_helper.py
eshiroma: Good call. Moved this comment to be part of the `Schema.validate()` docstring, and updated the `SchemaXServiceAccount.validate()` comment to reference it.

scripts/verify
eshiroma: Both `deploy.sh` [1] and `deploy_with_tests.sh` [2] already call `Schema.verify()` via `expand_config.py` [3], which. That's why the `schema.yaml` in our tests *had* to be updated despite no other new validation call additions, since the integration tests use the updated base deployer.

Happy to make other changes in a follow-up if this doesn't meet the concern.

[1] https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/master/marketplace/deployer_util/deploy.sh#L60
[2] https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/master/marketplace/deployer_util/deploy_with_tests.sh#L67
[3] https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/master/marketplace/deployer_util/expand_config.py#L69

PR #516, opened 2020-07-01
Remove dev image dependencies on schema validation

eshiroma:
To allow addition of new schema requirements without causing failures for existing apps. Warn instead of fully suppressing the validation failures to make users aware of the upcoming requirement.

Checked that `deploy.sh` and `deploy_with_tests.sh` covers all entrypoints (besides `bash`) from `dev` scripts, and therefore those which might invoke `config_helper`.

/gcbrun

### DIFF
--- a/marketplace/deployer_util/config_helper.py
+++ b/marketplace/deployer_util/config_helper.py
@@ -858,13 +858,6 @@ class SchemaXServiceAccount:
   """Accesses SERVICE_ACCOUNT property."""
 
   def __init__(self, dictionary):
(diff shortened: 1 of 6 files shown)


eshiroma:
All current calls to `validate()` for the schema are made in:
* `expand_config.py`
* `print_published_version.py`
* `print_version_metadata.py`

It is distinctly NOT used in `extract_schema_key`. Therefore, this is a no-op for `verify` and `install` (which only use `extract_schema_key`) and a tweak for `publish` (which uses print_published_version).

PR #519, opened 2020-07-20
Stop using 'cluster-admin' for deployer

eshiroma:
To improve security of deployment. Instead use `admin` and a custom `Role` for modifying Application CRs.

Re-submit #499 (rollback of #501), as existing applications which are expected to be broken by this change have been updated.

/gcbrun


eshiroma:
Dropping, since the corresponding UI deployment was not updated to use `admin`

PR #521, opened 2020-07-23
Disallow vague escalated permissions on deployerServiceAccount

eshiroma:
/gcbrun


marketplace/deployer_util/config_helper.py
```diff
@@ -168,6 +168,14 @@ def validate(self):
               'x-google-marketplace.images')
       if self._x_google_marketplace._deployer_service_account:
         self._x_google_marketplace._deployer_service_account.validate()
+        # Move to validate() once enforced on SERVICE_ACCOUNT properties as well.
+        if self._x_google_marketplace._deployer_service_account.has_discouraged_cluster_scoped_permissions(
```
huyhg, line 172: can be split into multiple lines with the following trick:
```python
if (self._x_google_marketplace
    ._deployer_service_account
    ._has_discouraged_cluster_scoped_permissions()):
```

eshiroma: Ah, much better - I forgot about that one since not added by our autoformatter.

eshiroma:
Not quite sure how the `.coverage` file was added in the first place (it's already checked in a HEAD despite `.coverage` being in the `.gitignore`), but I've removed it. Attempting to re-add it now properly triggers a gitignore warning.

---

/gcbrun

PR #524, opened 2020-08-03
Adding support and validation for assistedClusterCreation fields.

agarg2008:


marketplace/deployer_util/config_helper.py
```diff
@@ -531,6 +539,86 @@ def nodes(self):
...
+
+  def __init__(self, dictionary):
+    self._type = None
+    self._creation_guidance = None
+    self._gke = None
+
+    self._type = dictionary.get('type', None)
+    _must_contain(self._type, self._ASSISTED_CC_TYPES,
+                  "Invalid type of AssistedClusterCreation")
+    self._creation_guidance = dictionary.get('creationGuidance', "")
```
eshiroma, line 557: Use a default value of `None` to be consistent with our string default-value handling within this file.

marketplace/deployer_util/config_helper.py
```diff
@@ -531,6 +539,86 @@ def nodes(self):
...
+
+class SchemaNodePoolDetails:
+
+  def __init__(self, dictionary):
+    self._num_nodes = _must_get(dictionary, 'numNodes',
+                                'NodePoolDetails must have numNodes property')
+    self._machine_type = _must_get(
+        dictionary, 'machineType',
+        'NodePoolDetails must have machineType property')
+    if "custom-" in self._machine_type:
```
eshiroma, line 605: When you update `schema.md`, you should reference both the [predefined machine types](https://cloud.google.com/compute/docs/machine-types) and the [custom machine type docs](https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type#specifications) since there are too many requirements for us to document.

marketplace/deployer_util/config_helper.py
```diff
@@ -531,6 +539,86 @@ def nodes(self):
...
+      splits = re.split("-", self._machine_type)
+      cores = int(splits[-2])
```
eshiroma, line 607: Can you ensure there's no risk of `IndexError`?

marketplace/deployer_util/config_helper.py
agarg2008: Done.

marketplace/deployer_util/config_helper.py
agarg2008: Yes, will definitely do that.

marketplace/deployer_util/config_helper.py
agarg2008: Added check to ensure correct convention is used for custom-types.

PR #527, opened 2020-08-19
add assistedClusterCreation docs

agarg2008:
Adding assistedClusterCreation docs.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -923,7 +923,9 @@ indicates whether Istio is enabled on the cluster.
 
 ### assistedClusterCreation
 
-Use `assistedClusterCreation` to specify constraints for cluster creation as part of app deployment. These constraints determine whether a new cluster creation is allowed as part of deployment; and if yes, what are the node constraints.
+Use `assistedClusterCreation` to specify constraints for cluster creation as part of app deployment. These constraints determine whether a new cluster creation is allowed as part of deployment, and allow declaration of a specific node configuration for creation.
+
+`assistedClusterCreation` is optional if cluster creation is enabled. You can use it to specify a specific node configuration, but otherwise one will be generated according to defaults or specified `clusterConstraints`, if declared. The node config defined here will override the one suggested by `clusterConstraints`.
 
 ```yaml
 properties:
@@ -945,7 +947,7 @@ Supported types:
 - `DISABLED`: The app doesn't support creating a new cluster as part of app deployment. User must select a pre-existing cluster.
 - `STRICT`: The app specified strict requirements for number of nodes and machine type which should be used if creating a new cluster as part of app deployment. 
 
-Use `creationGuidance` to specify instructions which will explain to end-user how they should create a cluster that will be compatible with the app.
+Use `creationGuidance` to specify instructions which will explain to end-user how they should create a cluster that will be compatible with the app. `creationGuidance` is required if constraint type is set as `DISABLED`.
 
 For instance, the following specifies that a new cluster can not be created as part of deployment and customer must select an existing cluster with GPU.
 
@@ -958,7 +960,7 @@ x-google-marketplace:
   clusterConstraints:
     assistedClusterCreation:
       type: DISABLED
-      creationGuidance: "Please use existing cluster with GPU."
+      creationGuidance: "Create a cluster with Ubuntu as the node image type."
 ```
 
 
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -921,6 +921,68 @@ If this section is not specified, users see a warning when deploying the app to
 an Istio-enabled environment. The [`ISTIO_ENABLED`](#type-istio_enabled) type
 indicates whether Istio is enabled on the cluster.
 
+### assistedClusterCreation
+
+Use `assistedClusterCreation` to specify constraints for cluster creation as part of app deployment. These constraints determine whether a new cluster creation is allowed as part of deployment; and if yes, what are the node constraints.
```
eshiroma, line 926: "; and if yes, what are the node constraints." -->
", and allow declaration of a specific node configuration for creation."

The intro is probably a good place to call out that `assistedClusterCreation` is optional if cluster creation is enabled. Partners can use it to specify a specific node configuration, but otherwise one will be generated according to defaults or their specified `clusterConstraints`, if declared. The node config defined here will override the one suggested by `clusterConstraints`.

docs/schema.md
```diff
@@ -921,6 +921,68 @@ If this section is not specified, users see a warning when deploying the app to
...
+      type: DISABLED | STRICT
+      creationGuidance: #string
+      gke:
+        nodePool:
+        - numNodes: #integer
+          machineType: #string
+```
+
+Supported types:
+- `DISABLED`: The app doesn't support creating a new cluster as part of app deployment. User must select a pre-existing cluster.
```
eshiroma, line 945: Note here or in the `creationGuidance` explanatory line that `DISABLED` requires `creationGuidance`. We are fine with partners providing it when enabled, but we don't currently do anything with it and it is a bit awkward since Marketplace creates the cluster for the user (though we concieveably could show that info to help the user create their own customized cluster that will still work with the app)

docs/schema.md
```diff
@@ -921,6 +921,68 @@ If this section is not specified, users see a warning when deploying the app to
...
+```yaml
+properties:
+  # Property definitions...
+required:
+  # Required properties...
+x-google-marketplace:
+  clusterConstraints:
+    assistedClusterCreation:
+      type: DISABLED
+      creationGuidance: "Please use existing cluster with GPU."
```
eshiroma, line 961: (I know our internal tests use this line, but we already disable cluster creation automatically for GPUs (worth calling out in this section) so partners shouldn't use this feature for disabling creation when GPUs are needed, so can we use a different reason such as..."Create a cluster with Ubuntu as the node image type."?)

docs/schema.md
```diff
@@ -921,6 +921,68 @@ If this section is not specified, users see a warning when deploying the app to
...
+      type: STRICT
+      gke:
+        nodePool:
+        - numNodes: 2
+          machineType: custom-2-12288
+```
+`numNodes` is number of nodes to be created in each of the cluster's zones.
+`machineType` is the type of machine to use for nodes. It can either be a [predefined machine type](https://cloud.google.com/compute/docs/machine-types#general_purpose)  or [custom machine type](https://cloud.google.com/compute/docs/machine-types#custom_machine_types).
+
+Only `gke` clusters are supported at the moment.
```
eshiroma, line 984: " and a single `nodePool` definition"

docs/schema.md
```diff
@@ -921,6 +921,70 @@ If this section is not specified, users see a warning when deploying the app to
...
+          machineType: #string
+```
+
+Supported types:
+- `DISABLED`: The app doesn't support creating a new cluster as part of app deployment. User must select a pre-existing cluster.
+- `STRICT`: The app specified strict requirements for number of nodes and machine type which should be used if creating a new cluster as part of app deployment. 
+
+Use `creationGuidance` to specify instructions which will explain to end-user how they should create a cluster that will be compatible with the app. `creationGuidance` is required if constraint type is set as `DISABLED`.
+
+For instance, the following specifies that a new cluster can not be created as part of deployment and customer must select an existing cluster with GPU.
```
eshiroma, line 952: " and customer must select an existing cluster with GPU." ->
" and the user must create a new cluster with the Ubuntu node image."

docs/schema.md
```diff
@@ -921,6 +921,70 @@ If this section is not specified, users see a warning when deploying the app to
...
+  # Required properties...
+x-google-marketplace:
+  clusterConstraints:
+    assistedClusterCreation:
+      type: DISABLED
+      creationGuidance: "Create a cluster with Ubuntu as the node image type."
+```
+
+
+The following specifies that if a new cluster was to be created as part of deployment, it should be a new `gke` cluster with a `nodePool` containing 2 nodes of custom machine-type `custom-2-12288`.
```
eshiroma, line 967: Consider linking GKE cluster documentation here.

docs/schema.md
agarg2008: Done.

docs/schema.md
agarg2008: Synced offline and we decide not to link GKE cluster documentation at the moment.

PR #526, opened 2020-08-13
Fix schema validation warning

rstefan1:
fixes #525

PR #528, opened 2020-08-25
Fix load_schema_and_validate()

eshiroma:
/gcbrun

### DIFF
--- a/marketplace/deployer_util/schema_values_common.py
+++ b/marketplace/deployer_util/schema_values_common.py
@@ -62,7 +62,7 @@ def load_schema(parsed_args):
 
 
 def load_schema_and_validate(parsed_args):
-  return load_schema(parsed_args.schema_file).validate()
+  return load_schema(parsed_args).validate()
 
 
 @memoize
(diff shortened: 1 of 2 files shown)


gibbleyg:
While you're at it could you change this line to say something like "END OF WARNING"?

https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/blob/f6027496895acf7c6a179af86700768d4ece35bc/scripts/verify#L119

In the past I was confused and thought the stack trace here was the reason for verification failure.

gibbleyg:
Looks like there was already a PR for this as well https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/526

eshiroma:
Added `END OF WARNING`.

Ah - we should review issue/PR monitoring with ops.

/gcbrun

eshiroma:
For some reason, the cluster cleanup is failing (even on retry). Since all test steps (including `verify`) are successful, pushing.

eshiroma:
Nevermind - it finally succeeded. Possibly because I manually deleted the cluster.

PR #535, opened 2020-11-26
Add details to the sample UBB agent config

eshiroma:
Clarify the role of the `source` (`heartbeat`) section, and the responsibility of the app to report usage to the agent for all metrics except instance time.

/gcbrun


docs/billing-integration.md
```diff
@@ -82,6 +82,9 @@ data:
     # Metric names should match verbatim the identifiers created
     # during pricing setup.
     metrics:
+    # The app must send usage reports to the agent for this metric
+    # since it is not an instance time metric (which the agent can
```
gibbleyg, line 86: I think it's more precise to say that metrics are either generated by "native" sources or externally reported. Since this metric is not declared in the sources section, it must be externally reported

docs/billing-integration.md
```diff
@@ -138,6 +167,11 @@ data:
     # (defined by 'intervalSeconds'); this is assuming that the metric
     # configured in service control has 'seconds' as its unit.
     #
+    # This section is only a convenience for reporting instance
```
gibbleyg, line 170: Does this section just repeat the paragraph above? That heartbeat is the only supported metric?

docs/billing-integration.md
```diff
@@ -94,9 +97,35 @@ data:
       # The aggregation section indicates that reports that the agent
       # receives for this metric should be aggregated for a specified
       # period of time prior to being sent to the reporting endpoint.
+      # This is independent from how often the app reports usage to
```
gibbleyg, line 100: I'm confused, if the agent defers reporting until the metric has been aggregated for that period of time, how is it independent from how often the agent reports usage?

eshiroma: The `aggregation.bufferSeconds` is indeed tied to how often the *agent* reports usage - it defines it exactly. However, this is independent from how often the *app* reports usage to the agent.

Do you think it's better to leave off? The two kinds of reporting (app -> agent, agent -> google) didn't seem very clearly distinguished to partners, so I was trying to speak to that here.

gibbleyg: Got it, didn't understand the distinction being made between *app* and *agent* reporting. Up to you for what you think is best

Maybe change paragraph, "Metrics reported by the application are aggregated by the agent for a specified period of time and sent to the reporting endpoint (e.g. Google Service Control)"

gibbleyg (approved):

docs/billing-integration.md
eshiroma: Updated to:
```
    # Metric usage reports are either generated by the agent as
    # defined in "sources", or externally, i.e. by the app.
    # Since this metric is not supported by the "sources" section,
    # the app must send usage reports for this metric to the agent.
```

docs/billing-integration.md
eshiroma: Thanks - updated to the suggested paragraph.

docs/billing-integration.md
eshiroma: It does - removed an instead added to the first sentence in the prior paragraph to emphasize when this section should be omitted:

```
    # The sources section lists metric data sources run by the agent
    # itself - omit it if usage reports for all metrics are reported by
    # the application.
```

PR #536, opened 2020-12-02
Migrate from Travis to Github Actions

gibbleyg:


.github/workflows/ci-tests.yaml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run tests
+      run: make lint/py tests/py
```
huyhg, line 15: I don't recall exactly, but separating out the two targets for Travis might have made it run faster and perhaps the errors were easier to read. I don't think it matters thought--this is very fast.

huyhg (approved):

PR #545, opened 2021-02-09
Update schema docs for restricted SERVICE_ACCOUNT permissions

eshiroma:
Update docs now that `cluster-admin` is now disallowed in `SERVICE_ACCOUNT`properties. Recommend that `deployerServiceAccount` is used when long-lived service accounts are not needed.

/gcbrun


eshiroma:
To be submitted once #546 is in the latest dev image release, as this is when the stated restriction will go into effect.

vcanaa (approved):

PR #547, opened 2021-03-01
Adding instructions in building-deployer.md to handle CRD installation

vcanaa:


docs/building-deployer.md
```diff
@@ -107,3 +107,30 @@ re-published images should be named or tagged. This is because at deployment
...
 parameters. This is the reason why we require that all app images used by the
 deployer **must** be parameterized in `schema.yaml`.
+
+### Specific Scenarios
+
+#### Installing CustomResourceDefinitions (CRD)
+
+Solutions that include [CRDs](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
+templates as part of the deployment need to add extra permissions to the deployer
+service account in the schema.yaml file. Example:
```
eshiroma, line 117: nit: `schema.yaml`

docs/building-deployer.md
```diff
@@ -107,3 +107,30 @@ re-published images should be named or tagged. This is because at deployment
 time, full names of the app images are passed to the deployer as input
 parameters. This is the reason why we require that all app images used by the
 deployer **must** be parameterized in `schema.yaml`.
+
+### Specific Scenarios
+
+#### Installing CustomResourceDefinitions (CRD)
```
eshiroma, line 113: Should we maybe generalize as a preventative measure? e.g. "Installing Cluster-Scoped Resources"

CRDs are the most common example but it would be good to prevent a potential back-and-forth for the next partner with some other cluster-scoped resource.

(and updating the below to "Solutions that include cluster-scoped resources such as CRDs in the templates"...

eshiroma (approved):
Thank you for adding this!!

docs/building-deployer.md
vcanaa: I want partner to find the answer or the specific question, in the sense that someone does a text search on CRD and Custom resource and it points to this doc. If there are more specific examples, i dont mind adding them all one by one.

PR #548, opened 2021-03-03
Add CR instance to simple CRD example in tests

eshiroma:
To illustrate the CRD-before-CR requirement.

/gcbrun


eshiroma:
Retired in favor of other separate CRS (e.g. #549, #550, #551)

PR #550, opened 2021-03-04
Add example for creating CRD+CR with Helm

eshiroma:
Demonstrate creating a CRD+CR using the `deployer_helm_base`.

/gcbrun


eshiroma:
I believe the reason that this doesn't work as expected is that we aren't actually using `helm install`, we're using `helm template` and then applying manifests ourselves.
This does suggest that a change across our manifest-applying logic is needed but would work across both envsubst and helm (including helm2).

vcanaa (approved):

tests/marketplace/deployer_helm_base/onbuild/standard_v2/chart/minimal-helm-chart/Chart.yaml
```diff
@@ -3,3 +3,7 @@ appVersion: "1.0"
 description: A Minimal Helm 3 Chart
 name: minimal-helm-chart
 version: 0.1.0
+dependencies:
+- name: crd-chart
+  version: 0.1.0
+  repository: file://charts/crds
```
vcanaa, line 9: line end

vcanaa (approved):

huyhg:
#553 for reference. Our current deployer doesn't support CRD/CR ordering.

PR #562, opened 2021-06-03
Make app-crd compatible with CRD v1

gibbleyg:


crd/app-crd.yaml
```diff
@@ -11,202 +11,208 @@
...
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   creationTimestamp: null
   labels:
     controller-tools.k8s.io: "1.0"
   name: applications.app.k8s.io
+  annotations:
+    api-approved.kubernetes.io: https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/2
```
gibbleyg, line 22: @eshiroma not sure what we should do about this. I added this line for POC because the CRD is blocked by https://github.com/kubernetes/enhancements/pull/1111

I don't know if we could link back to the actual application sig, though there appears to be significant divergence https://github.com/kubernetes-sigs/application/blob/c8e2959e57a02b3877b394984a288f9178977d8b/config/crd/bases/app.k8s.io_applications.yaml#L8

gibbleyg: Here's an example of an annotation for an unapproved controller https://github.com/kubernetes/kubernetes/pull/101871#discussion_r631184970

It doesn't feel feasible to change Application to `v1alpha1` since the apiVersion is a parameter in `schema.yaml`

crd/app-crd.yaml
```diff
@@ -19,12 +19,13 @@ metadata:
     controller-tools.k8s.io: "1.0"
   name: applications.app.k8s.io
   annotations:
-    api-approved.kubernetes.io: https://github.com/GoogleCloudPlatform/marketplace-k8s-app-tools/pull/2
+    api-approved.kubernetes.io: unapproved, for Google Cloud Marketplace use only
```
gibbleyg, line 22: Following pattern in https://github.com/kubernetes/kubernetes/pull/101871#discussion_r631184970

crd/app-crd.yaml
```diff
@@ -19,12 +19,13 @@ metadata:
...
 spec:
   group: app.k8s.io
   names:
     kind: Application
     plural: applications
+    shortNames: [apps]
```
gibbleyg, line 28: Added shortname

eshiroma (approved):

PR #571, opened 2021-08-03
Remove schema docs only relevant to managed updates

eshiroma:
/gcbrun


eshiroma:
As published version info is no longer displayed anywhere in the UI.

gibbleyg (approved):

PR #572, opened 2021-08-04
Use WaitForFirstConsumer for GCP PersistentDisks

gibbleyg:
Pods can be unscheduable due to a Volume Node Affinity Conflict. This can happen when a pod has multiple pvcs, but the persistent volumes were created in different zones and thus not all pvs could be attached.

Kubernetes documentation suggests using WaitForFirstConsumer to avoid these issues. https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode

### DIFF
--- a/marketplace/deployer_util/provision.py
+++ b/marketplace/deployer_util/provision.py
@@ -671,9 +671,11 @@ def provision_storage_class(schema, prop, app_name, namespace, provisioner):
   if not provisioner:
     provisioner = _DEFAULT_STORAGE_CLASS_PROVISIONER
 
+  volume_binding_mode = ''
   if provisioner == 'kubernetes.io/vsphere-volume':
     parameters = {'diskformat': 'thin'}
   elif provisioner == 'kubernetes.io/gce-pd':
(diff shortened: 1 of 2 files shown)


marketplace/deployer_util/provision.py
```diff
@@ -671,9 +671,11 @@ def provision_storage_class(schema, prop, app_name, namespace, provisioner):
   if not provisioner:
     provisioner = _DEFAULT_STORAGE_CLASS_PROVISIONER
 
+  volume_binding_mode = 'Immediate'
```
eshiroma, line 674: Totally on board with making changes as non-invasive as possible. Though wondering if we have any reason to believe that the vsphere-volume wouldn't handle `WaitForFirstConsumer`? This branch might be hard to find later if we're trying to debug an Anthos-specific failure.

eshiroma (approved):

marketplace/deployer_util/provision.py
gibbleyg: This isn't supported for vsphere, will link to docs in code

PR #586, opened 2022-02-12
Creating multiple log files for mpdev verify

vcanaa:
Output from verify is very big and hard to consume. This change try to make troubleshooting easier by separating different sessions of the output in multiple files:

verify.log
The entire output. It includes the deployer logs

deployer.log
Only the log from deployer. It includes the tester logs.

tester.log
Only logs from the tester.

events.log
All the events collected before deleting the namespace

resources.yaml
All the resources that were applied. This session is huge and not often useful, so it is now removed from the output.

### DIFF
--- a/docs/mpdev-references.md
+++ b/docs/mpdev-references.md
@@ -123,18 +123,18 @@ values for verification.
 
 Logs from verify can be extensive, so they are separated in multiple files to facilitate troubleshooting
 
-- **resources.yaml:** Output of `kubectl get` for different kind of resources in the namespace. It should help visualize the footprint of the application after installation is complete.
+- `resources.yaml:` Output of `kubectl get` for different kind of resources in the namespace. It should help visualize the footprint of the application after installation is complete.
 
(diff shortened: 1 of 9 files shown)


scripts/dev
```diff
@@ -39,13 +39,18 @@ if [[ -t 0 ]]; then
   terminal_docker_param="-it"
 fi
 
+logdir="/usr/local/google/home/ruela/.mpdev_logs/$(date '+%Y%m%d-%H%M%S')"
```
vcanaa, line 42: Oops

docs/mpdev-references.md
```diff
@@ -119,6 +119,23 @@ Properties in `/data-test/schema.yaml` will override properties in
 `/data/schema.yaml`. This can also be used to overwrite existing default
 values for verification.
 
+#### Verification logs
+
+Logs from verify can be extensive, so they are separated in multiple files to facilitate troubleshooting
+
+- **resources.yaml:** Output of `kubectl get` for different kind of resources in the namespace. It should help visualize the footprint of the application after installation is complete.
```
eshiroma, line 126: nit: Can we wrap in ` for consistency with other filename references in the doc?

eshiroma (approved):

PR #635, opened 2025-03-18
Use "docker cp" to copy schema in install and verify script

cjsu-g:
Using docker cp without relying on the shell redirection to the local filesystem.


skywardpixel (approved):

kaiyangg (approved):
LGTM

PR #651, opened 2026-03-13
Update Dockerfile to use public ubuntu than a marketplace one

wutong327:
update ubuntu version to resolve b/487824300


erikcandeia (approved):
/gcbrun

PR #673, opened 2026-07-22
Fix race condition in weekly release

lamjonathan:
This PR adds a step to wait for the staging build to complete successfully before creating the tag and release, resolving a race condition where the prod build would try to pull non-existent staging images.

### DIFF
--- a/.github/workflows/create-weekly-image-update-release.yml
+++ b/.github/workflows/create-weekly-image-update-release.yml
@@ -19,6 +19,43 @@ jobs:
           fetch-depth: 0 # Fetch all history and tags
           token: ${{ secrets.PRODUCER_SAAP_BOT_PAT }}
 
+      - name: Wait for Staging Images Build
+        env:
+          GITHUB_TOKEN: ${{ secrets.PRODUCER_SAAP_BOT_PAT }}
+        run: |
+          SHA="${{ github.sha }}"
+          CHECK_NAME="Push-marketplace-k8s-app-tools (marketplace-k8s-app-tools)"
+          echo "Waiting for check '$CHECK_NAME' on commit $SHA to succeed..."
+          
+          max_attempts=30 # 30 minutes timeout
+          attempt=0
+          while [ $attempt -lt $max_attempts ]; do
+            # Query the check runs for the current commit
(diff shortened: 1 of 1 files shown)


lamjonathan:
# Explanation of Security Fixes in GHA Workflow
## 1. Pinning Actions to a Commit SHA (Line 17)
*   **Before**: `uses: actions/checkout@v4`
*   **After**: `uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7`
### Why this change was made:
*   GitHub tags (like `@v4`) are **mutable**. The maintainers of the action can update the tag to point to a different commit at any time (e.g., to release updates, or in a worst-case scenario, if the repository is compromised).
*   Pinning to a specific **immutable commit SHA** guarantees that your workflow always runs the exact same code that was audited and verified. This prevents "dependency hijacking" or supply chain attacks.
*   The SHA `692973e3d937129bcbf40652eb9f2f61becf3332` corresponds to the official release of `actions/checkout` version **`v4.1.7`**.
---
## 2. Resolving Template Injection Risks
`zizmor` flagged several lines where we used GitHub Action expression syntax (`${{ ... }}`) directly inside shell scripts.
### Affected Lines (Example):
*   **Before**:
    ```yaml
    run: |
      git tag "${{ steps.calculate_tag.outputs.NEW_TAG }}"
    ```
*   **After**:
    ```yaml
    env:
      NEW_TAG: ${{ steps.calculate_tag.outputs.NEW_TAG }}
    run: |
      git tag "$NEW_TAG"
    ```
### Why this change was made:
*   Using `${{ ... }}` directly inside a `run:` block causes GitHub to perform string substitution *before* the shell executes the script. If the substituted value contains shell metacharacters (such as `;`, `|`, `&`, or backticks), the shell might interpret them as commands, leading to **code injection**.
*   **The Fix**: By mapping the step output to an environment variable (`NEW_TAG` in the `env` block) and referencing it as a standard shell variable (`$NEW_TAG`), the value is treated strictly as data. The shell will not evaluate any metacharacters inside the variable, neutralizing the injection risk.
*   We applied this fix to the tagging, release, and staging-build-wait steps.

erikcandeia (approved):

PR #678, opened 2026-07-28
Fix public image tag rotation digest error

lamjonathan:
### Issue
The GCB tag rotation step "Manage Public Image Tags" failed with the following error:
`ERROR: (gcloud.container.images.add-tag) [gcr.io/...@ad02c056f254] digest must be of the form "sha256:<digest>".`

### Root Cause
The script was using `gcloud container images list-tags` to retrieve the digest of the old public image tag. However, `list-tags` truncates the digest to a 12-character short SHA by default. Even when using `--format="value(digest)"`, it returned the truncated value (e.g., `ad02c056f254`) without the required `sha256:` prefix. This truncated digest is invalid for the subsequent `add-tag` command.

### Proposed Fix & Justification
Replaced `gcloud container images list-tags` with `gcloud container images describe` to retrieve the digest of the specific old tag. 
Using `describe` with `--format="value(image_summary.digest)"` retrieves the full, untruncated 64-character SHA256 digest with the `sha256:` prefix, which is the correct format required by the `add-tag` command.

### DIFF
--- a/cloudbuild-tag.yaml
+++ b/cloudbuild-tag.yaml
@@ -62,8 +62,8 @@ steps:
       if [[ -n "$old_public_tag" ]]; then
         local old_version="${old_public_tag#public-image-}"
         local old_digest
-        old_digest=$(gcloud container images list-tags "$image_name" \
-          --filter="tags:$old_public_tag" --format="value(digest)")
+        old_digest=$(gcloud container images describe "$image_name:$old_public_tag" \
+          --format="value(image_summary.digest)")
         
         echo "[$image_name] Found old public image version: $old_version (digest: $old_digest)"
         


cloudbuild-tag.yaml
```diff
@@ -62,8 +62,8 @@ steps:
       if [[ -n "$old_public_tag" ]]; then
         local old_version="${old_public_tag#public-image-}"
         local old_digest
-        old_digest=$(gcloud container images list-tags "$image_name" \
-          --filter="tags:$old_public_tag" --format="value(digest)")
+        old_digest=$(gcloud container images describe "$image_name:$old_public_tag" \
+          --format="value(image_summary.digest)")
```
erikcandeia, line 66: Not sure I follow this fix. It seems like you changed the way you get the digest from `list-tags` to `describe`, but as I saw the screenshot, the issue was that the digest was missing the "sha256:" and the expected final format should be:
gcr.io/cloud-marketplace-tools/k8s/dev@sha256:<digest>
and currently it is:
gcr.io/cloud-marketplace-tools/k8s/dev@<digest>

lamjonathan: Yes, it seems that the old command to get the digest does not have the `sha256` prefix.

### Old Method (`list-tags`)
When running the old command to get the digest:
```bash
gcloud container images list-tags gcr.io/cloud-builders/gcloud --filter="tags:latest" --format="value(digest)"
```
It returns:
```
f0fb55d6de15
```
*   **Result:** Truncated to 12 characters and **missing** the `sha256:` prefix.
*   **Final command execution:** `gcloud container images add-tag gcr.io/cloud-builders/gcloud@f0fb55d6de15 ...` which fails because GCR does not accept truncated/prefix-less digests.


### New Method (`describe`)
When running the new command with `describe`:
```bash
gcloud container images describe gcr.io/cloud-builders/gcloud:latest --format="value(image_summary.digest)"
```
It returns:
```
sha256:f0fb55d6de1530ad854762fbaecf4d430ebad070cd2dc190e5061369267abf3b
```
*   **Result:** Full 64-character digest and **includes** the `sha256:` prefix.
*   **Final command execution:** `gcloud container images add-tag gcr.io/cloud-builders/gcloud@sha256:f0fb55d6de1530ad854762fbaecf4d430ebad070cd2dc190e5061369267abf3b ...` which is the correct format and will succeed.

erikcandeia (approved):

PR #691, opened 2026-09-17
Make integration test cluster creation resilient to GCE stockouts

cjsu-g:
## Problem

The `Initialize Credentials` step creates the ephemeral integration test cluster with `--zone ${_CLUSTER_LOCATION}`, and the triggers set that to `us-central1` — a **region**, not a zone. gcloud therefore creates a *regional* cluster with the default 3 nodes in **each** of 3 zones:

```yaml
location: us-central1
locations: [us-central1-f, us-central1-b, us-central1-c]
currentNodeCount: 9        # 9 x e2-medium = 18 vCPU, per build
```

Cluster creation is atomic, so the build fails if **any one** of those three zones is short on capacity at that moment. That is exactly what has been happening:

| Build | Result |
|---|---|
| `2081ed5e` (master @ `6f3747a`) | `GCE_STOCKOUT` / `UNAVAILABLE` — *"Google Compute Engine does not have enough resources available to fulfill request: us-central1"* |
| `f52afe85`, `a1ae54e9` (~2h earlier) | `GCE_QUOTA_EXCEEDED` — *"Quota 'CPUS' exceeded. Limit: 256.0 in region us-central1"* (~14 of these clusters saturate the regional quota) |
| `6bf3f7fd` | `TIMEOUT` — the 20 min build timeout expired while waiting on cluster creation |

The stockout on `2081ed5e` also failed the `Create Weekly Image Update Release` workflow, which blocks on this check.

Failures compounded the problem: `Cleanup` only runs when `Verify` succeeds, so a failed build left its cluster behind in `ERROR` state, still consuming CPU quota and making the next build more likely to fail.

## Changes

- **Retry across zones.** Candidate locations are tried in order, so a transient, location-specific error no longer fails the build. A *region* in `_CLUSTER_LOCATION` is expanded into its zones, so zonal clusters are created even with the current trigger configuration — no trigger change is required for this fix to take effect.
- **No leaked clusters from failed creation.** A failed creation leaves the cluster in `ERROR` state; it is now deleted before the next zone is tried.
- **Right-sized zonal cluster.** `_CLUSTER_NUM_NODES` nodes of `_CLUSTER_MACHINE_TYPE` (default 3 × `e2-standard-2`) in one zone, instead of 9 nodes spread over a region. `tests/integration` runs 4 small deployments; both values are substitutions so they are easy to tune if the tests ever need more room.
- **Cleanup is guaranteed.** `Verify` records its exit code instead of aborting the build, `Cleanup` always runs and deletes the cluster from the zone it was actually created in, and a new `Report verification result` step turns the recorded result back into the build status. Test failures are still build failures, and images are still only pushed when the build succeeds.
- **`_CLUSTER_LOCATION` gets a real default** instead of `null`, so builds that do not set the substitution stop passing an empty `--zone`.

## Testing

The step script was rendered exactly as Cloud Build would (`$$` → `$`, substitutions applied) and executed against stub `gcloud`/`kubectl` binaries that simulate stockouts:

| Scenario | Result |
|---|---|
| Preferred zone available | creates in `us-central1-c`, 1 create call, no cleanup |
| Stockout in preferred zone | falls back to `us-central1-f`, failed cluster deleted |
| Stockout in 3 zones | falls back to `us-east1-c`, all 3 failed clusters deleted |
| Region passed (current trigger config) | expands to zones, creates zonal cluster in `us-central1-b` |
| Every candidate stocked out | step fails with a clear message, all failed clusters deleted |

`bash -n` passes on every shell step, and the YAML parses with all anchors/aliases resolving.

> [!NOTE]
> The `_CLUSTER_LOCATION` substitution on the `Push-marketplace-k8s-app-tools` and `PR-marketplace-k8s-app-tools` triggers can optionally be changed from `us-central1` to a zone such as `us-central1-c` to skip the region-expansion lookup. It is not required — this change handles the region correctly on its own.


cjsu-g:
Validated end to end by this PR's own presubmit, build `3ce6601c-8826-4a97-b032-47f3e99b812a` (SUCCESS, all 10 steps green).

The trigger still passes `_CLUSTER_LOCATION=us-central1` (a region), and the new logic handled it correctly:

```
Step #6 - "Initialize Credentials": Creating cluster c-3ce6601c-... in us-central1-c ...
Step #6 - "Initialize Credentials": + CLUSTER_ZONE=us-central1-c
```

Cluster observed while the build was running — zonal, 3 nodes instead of 9 spread across a region:

```
NAME                                    LOCATION       STATUS   NODES
c-3ce6601c-8826-4a97-b032-47f3e99b812a  us-central1-c  RUNNING  3
```

`tests/integration` passed on the smaller cluster (~6 min, unchanged), and the cleanup handoff worked:

```
Step #8 - "Cleanup": ++ cat /workspace/cluster_zone
Step #8 - "Cleanup": + CLUSTER_ZONE=us-central1-c
Step #8 - "Cleanup": Deleted [.../zones/us-central1-c/clusters/c-3ce6601c-...].
Step #9 - "Report verification result": Finished
```

No cluster was leaked: `gcloud container clusters list` afterwards shows only the permanent `testing-marketplace` cluster.

Net effect per build: 9 VMs / 18 vCPU across 3 zones -> 3 VMs / 6 vCPU in 1 zone, with automatic failover to other zones when one is stocked out.

